PR #4, opened 2025-10-10
Clean up project structure

turtlebasket:
Originally forked this to add a tool for tailing the last *n* lines of logs for a pod, but realized this is unfortunately blocked by https://github.com/runpod/runpodctl/issues/29 ☹️ 

Anyway, a few fixes:

- build output (`build/*`) should not get tracked to git
- package-lock should not be ignored
- `.cursor/mcp.json` changed to use relative path and assume `RUNPOD_API_KEY` exported in Cursor environment
- more comprehensive gitignore


.cursor/mcp.json
```diff
@@ -1,10 +1,10 @@
 {
   "mcpServers": {
-    "runpod": {
+    "runpod-test": {
```
DeJayDev:
```suggestion
    "runpod": {
```

Was `-test` an artifact of local testing or is this to imply the user may have the production and a test copy of the MCP server running?

DeJayDev (commented):
Quick question, otherwise looks great - thanks you!

.cursor/mcp.json
turtlebasket: It's the latter- I had both installed so renamed the dev version for clarity.

TimPietruskyRunPod (approved):
lgtm! thank yo

PR #12, opened 2026-02-02
feat: add GPU/datacenter discovery tools and template filters

justinwlin:
## Summary

- **New tools**: `list-gpu-types` and `list-data-centers` using RunPod's public GraphQL API — allows LLMs to discover available hardware and regions without requiring authentication
- **Enhanced `list-templates`**: Added `includeRunpodTemplates`, `includePublicTemplates`, and `includeEndpointBoundTemplates` filter params (was previously a no-params tool returning only user templates)
- **Tool descriptions**: Added LLM-facing descriptions to `create-pod` and `list-templates` recommending **Runpod Pytorch 2.8.0** (ID: `runpod-torch-v280`) as the default template
- **README**: Added local build instructions for Claude Code CLI and Claude Desktop

### `list-gpu-types`
Queries the public GraphQL endpoint (`api.runpod.io/graphql`) — no API key needed. Supports filters:
- `minMemoryGb` — filter by VRAM (e.g., 80 for A100/H100)
- `secureCloudOnly` / `communityCloudOnly` — cloud type filter
- `searchTerm` — name search (e.g., "4090")
- `includeUnavailable` — include out-of-stock GPUs (excluded by default)

Results are sorted by stock availability then VRAM.

### `list-data-centers`
Also uses the public GraphQL endpoint. Supports `region` filter (e.g., "Europe", "United States").

## Test plan

- [x] `list-gpu-types` basic call returns GPUs sorted by stock/VRAM
- [x] `list-gpu-types` filters (minMemoryGb, secureCloudOnly, communityCloudOnly, searchTerm, includeUnavailable)
- [x] `list-data-centers` basic call and region filter
- [x] `list-templates` with each filter flag and combinations
- [x] Tool descriptions visible to LLM for `create-pod` and `list-templates`
- [x] Local build via `dist/index.mjs` connects and responds
- [x] `create-pod` functional test (requires real resources)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,6 +23,39 @@ const server = new McpServer({
   },
 });
 
+// GraphQL endpoint for public queries (GPU types, data centers)
+const GRAPHQL_URL = 'https://api.runpod.io/graphql';
+
+// Helper function to make GraphQL requests to RunPod
+async function graphqlRequest<T>(
+  query: string
+): Promise<T> {
+  const response = await fetch(GRAPHQL_URL, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ query }),
+  });
+
+  const result = (await response.json()) as {
+    data?: T;
+    errors?: Array<{ message: string }>;
+  };
+
+  if (result.errors && result.errors.length > 0) {
+    throw new Error(
+      `GraphQL Error: ${result.errors.map((e) => e.message).join(', ')}`
+    );
+  }
+
+  if (!result.data) {
+    throw new Error('No data returned from GraphQL query');
+  }
+
+  return result.data;
+}
+
 // Helper function to make authenticated API requests to RunPod
 async function runpodRequest(
   endpoint: string,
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -178,6 +383,7 @@ server.tool(
 // Create Pod
 server.tool(
   'create-pod',
+  'Create a new GPU/CPU pod on RunPod. If the user does not specify a template or image, recommend the "Runpod Pytorch 2.8.0" template (ID: runpod-torch-v280, image: runpod/pytorch:1.0.2-cu1281-torch280-ubuntu2404) as the default starting point — it has the most up-to-date CUDA and PyTorch versions.',
```
max4c: The new description tells the model to recommend the `runpod-torch-v280` template when the user does not specify a template or image, but the tool schema still requires `imageName` and does not expose `templateId`. That makes the newly documented template-based path impossible to execute through this tool and can push the model into invalid calls.

max4c (commented):
Review findings inline.

justinwlin:
Addressed the description mismatch:

- Removed the template ID reference (`runpod-torch-v280`) from the `create-pod` description. Since the tool only accepts `imageName` (no `templateId` field), the description now just recommends the image directly: `runpod/pytorch:1.0.2-cu1281-torch280-ubuntu2404`. The template ID mention was guiding the model toward a call it couldn't make.

justinwlin:
All feedback addressed. Validated against the live API.

**Line 386 — templateId in create-pod description**
Confirmed: the schema only has `imageName`, no `templateId` field. Removed the template ID reference from the description. It now recommends the image directly: `runpod/pytorch:1.0.2-cu1281-torch280-ubuntu2404`.

max4c (approved):

PR #13, opened 2026-02-02
feat: add Serverless endpoint runtime tools

justinwlin:
## Summary

- Add `serverlessRequest` helper targeting the Serverless API at `api.runpod.ai/v2`, separate from the REST management API.
- Add 8 new tools for invoking deployed Serverless workers: `run-endpoint`, `runsync-endpoint`, `get-job-status`, `stream-job`, `cancel-job`, `retry-job`, `endpoint-health`, and `purge-endpoint-queue`.
- Support optional `webhook`, `policy`, and `s3Config` parameters on run/runsync matching the full Serverless API spec.

## Test plan

- [ ] Build succeeds with `pnpm build`.
- [ ] `endpoint-health` returns worker counts and job stats for an active endpoint.
- [ ] `run-endpoint` submits async job and returns job ID with `IN_QUEUE` status.
- [ ] `get-job-status` returns `COMPLETED` with output after job finishes.
- [ ] `runsync-endpoint` returns completed output inline for a fast job.
- [ ] `cancel-job` cancels a queued job and returns `CANCELLED` status.
- [ ] `retry-job` requeues a failed job.
- [ ] `stream-job` returns incremental results from a streaming worker.
- [ ] `purge-endpoint-queue` clears pending jobs and returns removed count.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -743,49 +743,67 @@ server.tool(
 
 // ============== SERVERLESS RUNTIME TOOLS ==============
 
+// Shared schemas for serverless tools
+const endpointIdSchema = z
+  .string()
+  .regex(/^[a-zA-Z0-9_-]+$/, 'Invalid endpoint ID format');
+
+const jobIdSchema = z
+  .string()
+  .regex(/^[a-zA-Z0-9_-]+$/, 'Invalid job ID format');
+
+const inputSchema = z
+  .record(z.unknown())
+  .describe(
+    'Input payload for the worker handler. The expected fields depend on the deployed model or worker.'
+  );
+
+const webhookSchema = z
+  .string()
+  .url()
+  .optional()
+  .describe(
+    'Webhook URL to receive job completion notifications instead of polling'
+  );
+
+const policySchema = z
+  .object({
+    executionTimeout: z
+      .number()
+      .optional()
+      .describe('Maximum execution time in milliseconds'),
+    lowPriority: z
+      .boolean()
+      .optional()
+      .describe('Submit as a low-priority job'),
+    ttl: z
+      .number()
+      .optional()
+      .describe('Time-to-live for the job result in milliseconds'),
+  })
+  .optional()
+  .describe('Execution policy options');
+
+const s3ConfigSchema = z
+  .object({
+    accessId: z.string().describe('S3 access key ID'),
+    accessSecret: z.string().describe('S3 secret access key'),
+    bucketName: z.string().describe('S3 bucket name'),
+    endpointUrl: z.string().describe('S3 endpoint URL'),
+  })
+  .optional()
+  .describe('S3-compatible storage config for large outputs');
+
 // Run Endpoint (Async)
 server.tool(
   'run-endpoint',
   'Submit an asynchronous job to a Serverless endpoint. Returns a job ID immediately — use get-job-status to poll for results. Async results are available for 30 minutes after completion.',
   {
-    endpointId: z.string().describe('ID of the Serverless endpoint to run'),
-    input: z
-      .record(z.unknown())
-      .describe(
-        'Input payload for the worker handler. The expected fields depend on the deployed model or worker.'
-      ),
-    webhook: z
-      .string()
-      .optional()
-      .describe(
-        'Webhook URL to receive job completion notifications instead of polling'
-      ),
-    policy: z
-      .object({
-        executionTimeout: z
-          .number()
-          .optional()
-          .describe('Maximum execution time in milliseconds'),
-        lowPriority: z
-          .boolean()
-          .optional()
-          .describe('Submit as a low-priority job'),
-        ttl: z
-          .number()
-          .optional()
-          .describe('Time-to-live for the job result in milliseconds'),
-      })
-      .optional()
-      .describe('Execution policy options'),
-    s3Config: z
-      .object({
-        accessId: z.string().describe('S3 access key ID'),
-        accessSecret: z.string().describe('S3 secret access key'),
-        bucketName: z.string().describe('S3 bucket name'),
-        endpointUrl: z.string().describe('S3 endpoint URL'),
-      })
-      .optional()
-      .describe('S3-compatible storage config for large outputs'),
+    endpointId: endpointIdSchema.describe('ID of the Serverless endpoint to run'),
+    input: inputSchema,
+    webhook: webhookSchema,
+    policy: policySchema,
+    s3Config: s3ConfigSchema,
   },
   async (params) => {
     const { endpointId, ...body } = params;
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -488,6 +535,296 @@ server.tool(
...
+        },
+      ],
+    };
+  }
+);
+
+// Run Endpoint Sync
+server.tool(
+  'runsync-endpoint',
+  'Submit a synchronous job to a Serverless endpoint and wait for the result. Best for tasks completing within 90 seconds. If processing exceeds 90 seconds, the response returns a job ID to poll with get-job-status. Max payload: 20 MB. Results expire after 1 minute (up to 5 minutes with the wait parameter).',
```
max4c: The description says callers can extend synchronous waits with the `wait` parameter, but the schema exposes no `wait` input and the request always posts to `/runsync` with no query string. The official Runpod docs and SDK both support `?wait=...`, so jobs that need longer than the default wait window cannot use the behavior this tool claims to provide.

src/index.ts
```diff
@@ -488,6 +535,296 @@ server.tool(
...
+          text: JSON.stringify(result, null, 2),
+        },
+      ],
+    };
+  }
+);
+
+// Stream Job Results
+server.tool(
+  'stream-job',
```
max4c, line 895: `stream-job` calls `/stream/{jobId}` once and returns that one response. The official Runpod SDK treats `/stream` as a polling operation and loops until the job reaches a terminal state, yielding each `resp.stream` batch as it arrives. As written, long-running streamed jobs will return at most the current batch instead of the full stream the tool description promises.

max4c (commented):
Review findings inline.

max4c (commented):
Review findings inline.

justinwlin:
Addressed both comments:

1. **runsync wait param** — Removed the false claim about a `wait` parameter. The description now just says results expire after 1 minute, with no mention of an unimplemented extension.

2. **stream-job polling** — Replaced the single-shot call with a polling loop that calls `/stream/{jobId}` repeatedly (1s delay between polls), collecting all chunks into `allChunks[]`, until the status hits a terminal state (`COMPLETED`, `FAILED`, `CANCELLED`, or `TIMED_OUT`). The final response merges all collected chunks with the terminal result. Validated API key connectivity against the `16asy187ujq6bn` endpoint.

justinwlin:
Correction after live API testing:

**runsync wait param** — I was wrong to remove this. Tested against the live API and confirmed `?wait=` is a real server-side query param that takes milliseconds (valid range: 1000–300000). Previous fix deleted it; this commit properly adds `wait` to the schema (`z.number().min(1000).max(300000).optional()`) and passes it as `/runsync?wait=N` in the handler. The description now correctly documents the behavior.

**stream-job** fix stands — API confirmed the response shape is `{"status":"FAILED","stream":[]}`. The polling loop correctly checks for terminal status and collects all stream chunks.

justinwlin:
All feedback addressed. Validated against the live API (endpoint `16asy187ujq6bn`).

**Line 607 — runsync wait param**
Confirmed `?wait=` is a real server-side query param that takes milliseconds (valid range 1000–300000, tested — server returns 400 with "wait is invalid; must be between 1000 and 300000" on out-of-range values). Added `wait` to the schema as `z.number().min(1000).max(300000).optional()` and the handler now passes it as `/runsync?wait=N`. Description updated accordingly.

**Line 697 — stream-job single call**
Confirmed the response shape via live API: `{"status":"FAILED","stream":[]}`. Replaced the single call with a polling loop that calls `/stream/{jobId}` repeatedly (1s delay), accumulates all `stream[]` chunks, and stops when status hits `COMPLETED`, `FAILED`, `CANCELLED`, or `TIMED_OUT`.

justinwlin:
**Validation notes — how we determined the `wait` param**

Tested live against endpoint `16asy187ujq6bn` (Faster Whisper) using the API key.

**Confirming `wait` is real:**
```bash
# Out-of-range value — server rejects with 400
curl -X POST "https://api.runpod.ai/v2/16asy187ujq6bn/runsync?wait=29" ...
# → {"error":"wait is invalid; must be between 1000 and 300000"}

# Valid value — server accepts (HTTP 200), returns IN_QUEUE when workers are cold
curl -X POST "https://api.runpod.ai/v2/16asy187ujq6bn/runsync?wait=30000" ...
# → {"id":"sync-e737a2c6-...","status":"IN_QUEUE"}
```

The 400 error message gave us the exact valid range: **1000–300000 ms** (1 second to 5 minutes). Schema is `z.number().min(1000).max(300000).optional()`.

**Confirming stream response shape:**
```bash
curl "https://api.runpod.ai/v2/16asy187ujq6bn/stream/{jobId}" ...
# → {"status":"FAILED","stream":[]}
```

Confirmed the response has a top-level `status` field and a `stream` array — matching the polling loop implementation.

max4c (changes requested):
## Code Review

Overall solid PR — the 8 new serverless tools are well-structured and the previous feedback from @max4c has been addressed. However, there are two issues that should be fixed before merging:

### 1. 🔴 `stream-job` has an unbounded `while(true)` polling loop

The polling loop has no maximum iteration count, no overall timeout, and no error recovery for transient failures. If the API never returns a terminal status (network issues, stuck jobs, API bugs), this blocks the MCP server indefinitely.

```typescript
while (true) {
  const result = (await serverlessRequest(
    params.endpointId,
    `/stream/${params.jobId}`
  )) as Record<string, unknown>;
  // no max iterations, no timeout
  if (TERMINAL_STATUSES.has(result.status as string)) {
    break;
  }
  await new Promise((resolve) => setTimeout(resolve, 1000));
}
```

**Fix:** Add a max poll duration (e.g., 10 minutes) or iteration cap, and return whatever has been collected so far with a timeout indicator. Also wrap the `serverlessRequest` call in try/catch so a transient 429/502 doesn't crash the tool and discard all previously collected chunks.

### 2. 🟡 `endpointId` interpolated into URL without validation

`endpointId` and `jobId` are `z.string()` with no format constraints. A value like `../../v1/other-endpoint` would manipulate the URL path:

```typescript
const url = `${SERVERLESS_API_BASE_URL}/${endpointId}${path}`;
```

**Fix:** Add a regex constraint:
```typescript
endpointId: z.string().regex(/^[a-zA-Z0-9_-]+$/).describe(...)
```
Or at minimum use `encodeURIComponent()`.

### 3. 🟡 Duplicated schemas & minor validation gaps

- The `policy`, `s3Config`, `webhook`, and `input` Zod schemas are duplicated verbatim between `run-endpoint` and `runsync-endpoint`. Extract into shared constants.
- `webhook` should be `z.string().url().optional()` instead of `z.string().optional()`.

---

Items 1 and 2 should be addressed before merge. Item 3 is a nice-to-have.

justinwlin:
Addressed all three items from the latest review:

1. **`stream-job` unbounded polling** — Added 5-minute timeout and try/catch with a 5-consecutive-error abort. Collected chunks are preserved on timeout or error.
2. **`endpointId`/`jobId` path traversal** — Added `z.string().regex(/^[a-zA-Z0-9_-]+$/)` validation on both, applied across all 8 serverless tools.
3. **Duplicated schemas** — Extracted shared constants (`endpointIdSchema`, `jobIdSchema`, `inputSchema`, `webhookSchema`, `policySchema`, `s3ConfigSchema`). `webhook` now validates with `.url()`.

max4c (approved):

PR #14, opened 2026-02-13
Add .arch/repo.yaml for architecture data mesh

justinwlin:
## Summary

- Adds `.arch/repo.yaml` to register this repo in the RunPod architecture data mesh
- Auto-discovered by `build-db.ts` in [runpod/architecture](https://github.com/runpod/architecture)
- Includes: 7 handlers, 1 env var, 4 conventions, 3 failure modes, 2 key files

Companion PR: https://github.com/runpod/architecture/pull/27

PR #16, opened 2026-03-06
fix: correct .arch/repo.yaml service_calls and conventions

justinwlin:
Fixes errors in the `.arch/repo.yaml` merged in #14.

## Changes
- **Remove fabricated `serverless-runtime` service_call** — `api.runpod.ai/v2` does not exist anywhere in the codebase. All tools use a single base URL: `https://rest.runpod.io/v1` via `runpodRequest()`
- **Fix conventions** — updated from "Two API helpers" to correctly state there is one helper (`runpodRequest()`) and one base URL

### DIFF
--- a/.arch/repo.yaml
+++ b/.arch/repo.yaml
@@ -78,14 +78,9 @@ service_calls:
   - target: rest-api
     type: http
     protocol: https
-    purpose: All Runpod REST API operations (rest.runpod.io/v1)
+    purpose: All Runpod REST API operations including pods, endpoints, templates, network volumes, container registry, and serverless job management — all via https://rest.runpod.io/v1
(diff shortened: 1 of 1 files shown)


TimPietruskyRunPod (approved):
Looks good — correctly removes the fabricated serverless-runtime service call and consolidates to the single runpodRequest() helper. search_pattern addition is a nice touch for discoverability.

max4c (commented):
The service-call section is corrected, but the metadata still advertises runtime handlers like `run-endpoint`, `runsync-endpoint`, `get-job-status`, and `stream-job` that do not exist on current `main`. Because this PR is metadata-only, that stale handler inventory is the main remaining correctness issue.

PR #24, opened 2026-06-03
feat: send user-agent header on outbound runpod api requests

### LINKED ISSUE #23: user-agent in runpodRequest function
To better serve agents, we'd like to identify requests that were assisted with the Runpod MCP server. Can we add a User-Agent header in index.tx in the runpodRequest function?

TimPietruskyRunPod:
## Summary

- Adds `User-Agent: runpod-mcp-server/<version>` to every outbound request from the MCP server: `runpodRequest` (REST), `serverlessRequest` (Serverless runtime), and `graphqlRequest` (public GraphQL). This lets API logs attribute MCP-driven traffic.
- The version is read from `package.json` at build time via tsup's `define`, so it stays in sync with `changeset publish` automatically. A `dev` fallback is used when running through `tsx` (e.g. `pnpm dev`), producing `runpod-mcp-server/dev`.
- Includes a small amount of prettier reformatting in `src/index.ts` (long `.describe()` strings and one `JSON.stringify` block) so `pnpm prettier-check` passes on the branch — those lines already failed the project's prettier config on `main`.

Closes #23

## Test plan

- [x] `pnpm type-check` passes.
- [x] `pnpm build` succeeds with no warnings; verified `"1.1.0"` is inlined into both `dist/index.mjs` and `dist/index.js`.
- [x] `User-Agent: USER_AGENT` appears at 3 call sites in each bundled output (one per helper).
- [x] `pnpm lint` and `pnpm prettier-check` pass.
- [ ] Manual: run the built server and confirm the header reaches the Runpod API logs as `runpod-mcp-server/1.1.0`.

### DIFF
--- a/.changeset/user-agent-header.md
+++ b/.changeset/user-agent-header.md
@@ -1,0 +1,5 @@
+---
+'@runpod/mcp-server': patch
+---
+
+Send a `User-Agent: runpod-mcp-server/<version>` header on every outbound request to the Runpod REST, Serverless, and GraphQL APIs so MCP-driven traffic can be attributed in API logs.
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,11 @@ import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js'
 import { z } from 'zod';
 import fetch, { type RequestInit as NodeFetchRequestInit } from 'node-fetch';
 
+declare const __PACKAGE_VERSION__: string;
+const PACKAGE_VERSION =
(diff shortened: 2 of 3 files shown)


TimPietruskyRunPod:
Closing in favor of #25 — Luke's approach is strictly better: it identifies the actual MCP client (Claude Code, Cursor, Codex, etc.) via the `initialize` handshake's `clientInfo`, which is what #23 actually wants. This PR only said "some MCP server made the call," which is a subset of what #25 already does.

PR #25, opened 2026-06-03
feat: add caller tracking headers to all outbound API calls

lukepiette:
Every outbound HTTP request now carries a structured User-Agent identifying the calling MCP client and an anonymous per-process X-Runpod-Session-Id header so the Runpod platform can attribute traffic to specific MCP clients (Claude Code, Cursor, Codex, Gemini CLI, etc.) and count distinct agent sessions.

The User-Agent format is:

  runpod-mcp-server/{version} (caller=mcp; client=<name>; client_version=<ver>; transport=<stdio|http>)

Client identity comes from the MCP `initialize` handshake's `clientInfo`, read via `server.server.getClientVersion()`. The MCP spec requires every client to send this, so it covers 100% of MCP clients in the wild.

The transport token is set at runtime via `setTransport('stdio'|'http')`; stdio entrypoints get the default, and the Streamable HTTP entrypoint (`feat/http-transport`) will call `setTransport('http')` before booting the server.

Pure observability — no tool behavior, parameter, or return shape changes. The three HTTP helpers (runpodRequest, graphqlRequest, serverlessRequest) inject the new headers automatically.

Test plan:
- Type check passes (tsc --build).
- Build succeeds (tsup).
- Lint clean (eslint).
- Prettier formatting verified.
- Manual smoke test with a stub MCP client to confirm the User-Agent reflects whichever client connected is recommended before release.


src/index.ts
```diff
@@ -13,27 +14,94 @@ if (!API_KEY) {
   process.exit(1);
 }
 
+// Server identity — also used in the outbound User-Agent header.
+const MCP_SERVER_NAME = 'RunPod API Server';
+const MCP_SERVER_VERSION = '1.2.0';
```
TimPietruskyRunPod: **Hardcoded version drifts from `package.json`.** This PR sets `MCP_SERVER_VERSION = '1.2.0'`, but `package.json` is still at `1.1.0` — so the UA is already lying on `main`. Every future release will require remembering to bump both this constant and `package.json`, which is a foot-gun waiting to happen (this exact bug ships the day someone forgets).

Suggest reading from `package.json` at build time via tsup's `define` so it always matches the published version automatically (works with `changeset publish`):

```ts
// tsup.config.ts
import { createRequire } from 'node:module';
import { defineConfig } from 'tsup';

const { version } = createRequire(import.meta.url)('./package.json') as {
  version: string;
};

export default defineConfig([{
  // ...existing config...
  define: {
    __PACKAGE_VERSION__: JSON.stringify(version),
  },
}]);
```

```ts
// src/index.ts
declare const __PACKAGE_VERSION__: string;
const MCP_SERVER_VERSION =
  typeof __PACKAGE_VERSION__ !== 'undefined' ? __PACKAGE_VERSION__ : 'dev';
```

The `'dev'` fallback covers `pnpm dev` (tsx) where tsup's `define` doesn't apply — produces a useful `runpod-mcp-server/dev` UA in local development.

src/index.ts
```diff
@@ -13,27 +14,94 @@ if (!API_KEY) {
...
+ *   runpod-mcp-server/<version> (caller=mcp; client=<name>; client_version=<ver>; transport=stdio)
+ *
+ * Gateway parsers should treat unknown tokens as opaque and key off the
+ * `caller=` / `client=` / `client_version=` / `transport=` pairs.
+ */
+function buildUserAgent(): string {
+  const { name, version } = resolveClientInfo();
+  const safeName = sanitizeUaToken(name);
+  const safeVersion = sanitizeUaToken(version);
+  return `runpod-mcp-server/${MCP_SERVER_VERSION} (caller=mcp; client=${safeName}; client_version=${safeVersion}; transport=stdio)`;
```
TimPietruskyRunPod: **`transport=stdio` is hardcoded.** Fine on `main` today, but the `feat/http-transport` branch (in active development) adds a Streamable HTTP transport — once that lands and gets merged, every HTTP call will go out claiming `transport=stdio`, which defeats the point of having the field.

Two low-friction options:

1. Module-level variable that the entrypoint sets:

```ts
let MCP_TRANSPORT: 'stdio' | 'http' = 'stdio';
export function setTransport(t: 'stdio' | 'http') { MCP_TRANSPORT = t; }
```

Then call `setTransport('http')` from the HTTP entrypoint when that wiring lands.

2. Or read from an env var (`RUNPOD_MCP_TRANSPORT`) set per entrypoint — zero coupling.

Either way, this needs to be dynamic before the HTTP transport ships, otherwise the data side will undercount HTTP traffic.

src/index.ts
```diff
@@ -13,27 +14,94 @@ if (!API_KEY) {
...
+}
+
+/**
+ * Headers to attach to every outbound HTTP call so the Runpod platform can
+ * attribute traffic to an MCP session and client.
+ */
+function trackingHeaders(): Record<string, string> {
+  return {
+    'User-Agent': buildUserAgent(),
+    'X-Runpod-Session-Id': SESSION_ID,
```
TimPietruskyRunPod, line 106: Sanity-check: does the Runpod gateway actually consume `X-Runpod-Session-Id` server-side today (across REST `rest.runpod.io`, Serverless `api.runpod.ai/v2`, and GraphQL `api.runpod.io/graphql`)? If nothing parses or logs it yet, we're shipping a 36-byte header on every outbound request for nothing.

Not blocking the design — the anonymous-per-process UUID is a clean approach — but worth a +1 from whoever owns the gateway/data side that they're ready to receive and aggregate on it. Otherwise this should land alongside the consumer, not ahead of it.

(Same question applies to the structured `caller=`/`client=`/`client_version=`/`transport=` tokens in the UA — confirming the parsing logic exists or is planned would be helpful.)

TimPietruskyRunPod (changes requested):
Closing my own PR #24 in favor of this — the design is strictly right. Sourcing client identity from the `initialize` handshake's `clientInfo` is the canonical, spec-compliant approach, and structured tokens + sanitization + length cap + anonymous per-process session ID are all good calls. Pure observability with no tool behavior changes is exactly the scope that issue #23 asks for.

Three blocking items before merge:

1. **`MCP_SERVER_VERSION` is hardcoded** to `'1.2.0'` while `package.json` is at `1.1.0`. It's already drifted in this PR, and will drift again every release unless someone hand-bumps both. Inline suggestion below to read from `package.json` at build time.
2. **`transport=stdio` is hardcoded** in the UA. The `feat/http-transport` branch (active development) adds a Streamable HTTP transport — once that lands, the UA on HTTP calls will incorrectly report `stdio`. Inline suggestion below.
3. **`X-Runpod-Session-Id`** — does the Runpod gateway actually consume this header server-side today (REST, Serverless, GraphQL), or is this anticipatory? If nothing logs or parses it yet, we're shipping a header for nothing. Worth a confirmation from whoever owns the gateway/data side before merging.

Everything else looks good — happy to re-approve once these are addressed.

.changeset/caller-tracking.md
```diff
@@ -0,0 +1,5 @@
+---
+"@runpod/mcp-server": minor
+---
+
+Add caller tracking headers to every outbound API call. Each request now carries a structured `User-Agent` (`runpod-mcp-server/<version> (caller=mcp; client=<mcp_client_name>; client_version=<mcp_client_version>; transport=stdio)`) and an anonymous per-process `X-Runpod-Session-Id` UUID. The MCP client identity is sourced from the `initialize` handshake's `clientInfo`. No tool behavior changes — observability only — so the Runpod platform can attribute traffic to specific MCP clients (Claude Code, Cursor, Codex, Gemini CLI, etc.) and count distinct agent sessions.
```
TimPietruskyRunPod: Tiny doc-consistency nit: the example UA still says `transport=stdio` literally, but since `50f9a24` introduced `setTransport()` the value is dynamic. Suggest matching the source-code comment on line 87 of `src/index.ts`:

```suggestion
Add caller tracking headers to every outbound API call. Each request now carries a structured `User-Agent` (`runpod-mcp-server/<version> (caller=mcp; client=<mcp_client_name>; client_version=<mcp_client_version>; transport=<stdio|http>)`) and an anonymous per-process `X-Runpod-Session-Id` UUID. The MCP client identity is sourced from the `initialize` handshake's `clientInfo`. No tool behavior changes — observability only — so the Runpod platform can attribute traffic to specific MCP clients (Claude Code, Cursor, Codex, Gemini CLI, etc.) and count distinct agent sessions.
```

(Same fix applies to the PR description body — but that's not a file the reviewer tool can suggest into.)

TimPietruskyRunPod (approved):
Re-reviewing after `50f9a24` — code blockers #1 and #2 from my previous review are fully addressed. Adopted the tsup `define` + `'dev'` fallback exactly, and `setTransport('stdio'|'http')` is a clean module-level toggle that the `feat/http-transport` entrypoint can call before booting the server.

Nice bonus: `new McpServer({ version: MCP_SERVER_VERSION })` now reports the real `package.json` version on the `initialize` handshake too, instead of the stale hardcoded `'1.0.0'`. Free win.

Approving the code. Two non-code items still open before merge:

1. **`X-Runpod-Session-Id` gateway consumption** (line 106) — still unanswered. Please confirm with whoever owns the gateway/data side that REST (`rest.runpod.io`), Serverless (`api.runpod.ai/v2`), and GraphQL (`api.runpod.io/graphql`) ingest pipelines actually parse/log this header (and ideally the structured UA tokens) before merging. If not, this PR should ship alongside that consumer work, not ahead of it — otherwise we're shipping a header into the void and won't catch regressions if the format drifts.

2. **Doc nit (non-blocking)**: the changeset (`.changeset/caller-tracking.md`) and PR description still show `transport=stdio` literally in the example User-Agent. Since the value is now dynamic, both should say `transport=<stdio|http>` to match. Inline suggestion below.

TimPietruskyRunPod:
Pushed the doc nit directly as `80f4b01` so you don't have to round-trip — just the changeset `transport=stdio` → `transport=<stdio|http>` fix from my last review, plus the matching one-line update in the PR description. Approval still stands. Only outstanding item is the gateway-side confirmation on `X-Runpod-Session-Id` consumption before merge.

PR #27, opened 2026-06-12
feat: hosted HTTP transport + Sign in with Runpod (OAuth) for the MCP server

TimPietruskyRunPod:
Linear: [DR-1352 — Clerk OAuth app for Runpod remote MCP server v2 auth](https://linear.app/runpod/issue/DR-1352/clerk-oauth-app-for-runpod-remote-mcp-server-v2-auth)

## Summary

Adds a hosted **Streamable HTTP** transport for the Runpod MCP server alongside the existing local `stdio` mode, plus a full **"Sign in with Runpod"** OAuth flow so remote clients (Claude Desktop/Code) can authenticate without a pre-shared key.

The hosted server is **stateless token passthrough**: each request carries its own `Authorization: Bearer <token>`, which the server forwards to the Runpod API and then discards. It holds no credential of its own — no shared key.

### What's included
- **Source split**: `src/stdio.ts`, `src/http.ts`, `src/server.ts`, `src/tools.ts`; three package entrypoints (`.`, `./http`, `./tools`); Vercel adapter in `api/index.ts`.
- **OAuth authorization server** (enabled via `MCP_OAUTH_ENABLED=true`): `/.well-known/*` discovery, `POST /register` (RFC 7591 dynamic client registration), `GET /authorize`, `POST /token`.
- **Flash-backed flow that mints a real Runpod API key**: `/authorize` → `createFlashAuthRequest` → console approval (`/integrations/mcp/login`) → `/token` polls `flashAuthRequestStatus` → returns the minted `apiKey` as the access token. The client (Claude) stores the key; the MCP never persists it.
- **Optional key naming**: `RUNPOD_API_KEY_NAME` passes `apiKeyName` to the backend so the minted key is named (e.g. `runpod-mcp`) in the dashboard. Omitted by default for compatibility with backends that haven't shipped the argument.
- **Configurable hosts**: `RUNPOD_GRAPHQL_URL`, `RUNPOD_REST_API_URL`, `RUNPOD_SERVERLESS_API_URL`, `CONSOLE_BASE_URL` for targeting non-production environments.
- **Caller-tracking preserved**: the `User-Agent` + `X-Runpod-Session-Id` observability headers from `main` (#25) are ported into the split modules and report the transport (`stdio`/`http`).
- **`scripts/oauth-e2e.ts`**: a debug harness that drives the entire flow as a real MCP client (discovery → register → authorize → you approve in a browser → poll `/token` → connect + call a tool). Talks only to the deployment — no backend URL or credential needed.

### Environment variables (hosted/OAuth)
| Var | Purpose |
|---|---|
| `MCP_OAUTH_ENABLED` | `true` to advertise the OAuth sign-in flow |
| `RUNPOD_GRAPHQL_URL` | flash auth backend (default `https://api.runpod.io/graphql`) |
| `CONSOLE_BASE_URL` | console hosting the handoff page (default `http://localhost:3000`) |
| `RUNPOD_REST_API_URL` / `RUNPOD_SERVERLESS_API_URL` | override API hosts for non-prod keys |
| `RUNPOD_API_KEY_NAME` | optional name for the minted key (only send to a backend that supports `apiKeyName`) |

## Test plan (verified)
- `stdio` and HTTP smoke tests pass against the live API.
- **Full OAuth chain verified end-to-end** against a Vercel preview: discovery → DCR → `/authorize` → browser approval → console redirect back to `redirect_uri?code=…` → `/token` returned a minted `rpa_…` key → MCP connected with it.
- **Key naming verified**: with `RUNPOD_API_KEY_NAME=runpod-mcp` against a backend exposing `apiKeyName`, the minted key is named accordingly.
- Rebased onto latest `main`; CI green on Node 18/20/22; type-check, lint, prettier, build all pass.

## Notes
- **PKCE** is advertised (`S256`) but not enforced server-side (the flash flow has no PKCE); security relies on the single-use, short-lived flash request + console approval.
- **`apiKeyName` rollout**: the argument must be deployed on the target backend before setting `RUNPOD_API_KEY_NAME` there — sending it to a backend without it fails with "Unknown argument". This is why it's opt-in and omitted by default.


justinwlin:
## Review — 🔴 High (1/3)

Did an AI-assisted review pass (tool parity, OAuth security, build/packaging, runtime edge-cases). Overall the refactor is clean — all 34 tools preserved byte-for-byte, npm publish + `npx` install verified intact, no dropped functionality. One high-severity item worth resolving before merge, plus a batch of medium/low in the follow-up comments.

### `redirect_uri` is never validated — authorization code (→ real API key) can be exfiltrated

**Where:** `api/index.ts` — `/register` (~L171), `/authorize` (~L207), `/token` (~L262)

`/register` accepts arbitrary `redirect_uris` and persists nothing. `/authorize` reads `redirect_uri` straight from the query with no allowlist check. `/token` reads `redirect_uri` only to log it — never compares it. Because the authorization code redeems at `/token` into a **minted real, long-lived Runpod API key**, whoever receives the redirect callback gets the code, and whoever redeems the code gets a live key on the victim's account.

**Attack sketch:**
1. Attacker crafts an `/authorize` link with `redirect_uri=https://attacker.example/cb` (legit Runpod AS, real console login page).
2. Victim approves in good faith — nothing on a genuine "Sign in with Runpod" screen tells them where the credential will be delivered.
3. Console redirects to `https://attacker.example/cb?code=<id>`.
4. Attacker POSTs the code to `/token`, receives a real `rpa_…` key on the victim's account.

The interactive console approval does **not** mitigate this — the redirect destination is the only thing deciding who receives the credential, and it's unchecked. Classic auth-code interception, made worse because the code redeems to a permanent full-privilege key rather than a scoped/expiring token.

**Caveat to confirm:** part of the redirect happens on the console side (`/integrations/mcp/login`), outside this repo. If the console independently validates `redirect_uri`, real-world severity drops — worth confirming with whoever owns the handoff page. The MCP server itself provides no validation and shouldn't assume the console does.

**Suggested fix:** persist `redirect_uris` at `/register`, map `client_id` → allowed URIs, exact-match `redirect_uri` at `/authorize`, and re-validate at `/token` (requires binding the code to it — see PKCE/code-binding items in the Medium comment). At minimum, enforce an exact-match allowlist.

justinwlin:
## Review — 🟡 Medium (2/3)

### 1. PKCE advertised (S256) but not enforced
`api/index.ts` discovery advertises `code_challenge_methods_supported: ['S256']`, but `code_challenge` is only captured as a log boolean at `/authorize` and `code_verifier` is never verified at `/token`. PKCE's whole job is to prevent the code interception in the High comment — advertising it while not enforcing it is a downgrade trap (clients believe they're protected). Either enforce it (store challenge bound to the flash request, verify `SHA256(verifier)` at `/token`) or stop advertising it.

### 2. Authorization code not bound to client / redirect_uri
`/token` uses `token_endpoint_auth_methods: ['none']` and authenticates nothing — redemption depends only on possessing the code. No PKCE, no client secret, no redirect_uri match. Single-use is the only defense, and it's fully delegated to the flash backend (`CONSUMED` + null key → "already used"). `handleToken` polls up to 6× and would return the key on each successful poll until the backend flips status — so replay safety depends entirely on the backend atomically consuming the key on first read. Worth confirming + documenting that guarantee.

### 3. OAuth routes are always live, ignoring `MCP_OAUTH_ENABLED`
`MCP_OAUTH_ENABLED` (in `src/http.ts:isOAuthEnabled`) only gates the `WWW-Authenticate` 401 challenge hint. The actual routes in `api/index.ts` (`/.well-known/*`, `/authorize`, `/register`, `/token`) have no gating check — they respond on every deployment. So the High-comment attack surface exists even on deploys where an operator believes OAuth is "disabled." Suggest gating the route handlers behind the same `isOAuthEnabled()` (return 404 when off) and sharing one helper.

### 4. Access token is a permanent, full-privilege API key
`/token` returns `access_token: status.apiKey` with no `expires_in`, no `refresh_token`, no `scope`, and no `revocation_endpoint` advertised. The client (Claude) stores a permanent full-privilege Runpod credential; compromise of its token store = account compromise until manually revoked. `RUNPOD_API_KEY_NAME` is opt-in, so by default minted keys are hard to identify in the dashboard for revocation. Prefer scoped/expiring keys + `expires_in`; always name minted keys for the hosted flow; document the long-lived-credential risk.

### 5. `CONSOLE_BASE_URL` defaults to `http://localhost:3000`
`getConsoleBaseUrl()` defaults to localhost, so a prod deploy that forgets `CONSOLE_BASE_URL` 302s every user to `http://localhost:3000/login`. Silent (discovery/authorize still return 200/302). Contrast `getRunpodGraphqlUrl()`, which safely defaults to production. Default to the prod console URL, or fail closed if unset in a non-dev env.

### 6. Caller-tracking reports `client=unknown` over hosted HTTP
In stateless HTTP mode a fresh `McpServer` is created per request (`src/http.ts:111`). MCP names the client during `initialize`, but a `tools/call` is a separate request with its own fresh server that never saw that `initialize` — so `server.getClientVersion()` (`src/tools.ts:54`) returns undefined and the `User-Agent` reports `client=unknown` on essentially every hosted tool call. The ported #25 attribution works for stdio (long-lived server) but silently fails for the hosted path. (Compounds with the `version=dev` low item.)

### 7. `/token` returns `authorization_pending` as HTTP 400 — non-retryable, can dead-end
`/token` polls ~6×2s (~10s) then, if still PENDING, returns `400 {error:"authorization_pending"}`. That's a device-flow (RFC 8628) "keep polling" signal, but this is a browser-redirect `authorization_code` grant — a standard client (Claude's connector) typically treats a token-endpoint 400 as terminal and won't re-poll. Usually masked (status is APPROVED on first poll), but if console→backend propagation lag exceeds ~10s the user hits a hard failure with a stranded code and must restart. Timing isn't the risk (10s ≪ 60s maxDuration) — the non-retryable 400 is. Worth confirming Claude's client behavior here.

### 8. `.changeset/` is gitignored but changesets are committed + required by releases
`.gitignore:3` has `.changeset/`, yet 6 changeset files are tracked (force-added). A contributor following the documented "add a changeset" workflow will create `.changeset/foo.md`, see it absent from `git status`, and never commit it — quietly breaking the next release's changelog/version bump. Likely pre-existing, but this PR adds 5 changesets and leans on the workflow. Remove the `.changeset/` line (changesets are source, meant to be tracked).

justinwlin:
## Review — 🟢 Low / Nit (3/3)

**Note on two things we checked and cleared:** an automated pass initially flagged the stateless HTTP transport as broken on Vercel (SSE buffering; `handleRequest` called without the parsed body). We verified against the installed SDK (v1.20.2) + Vercel runtime + your e2e evidence — **not broken**: Vercel supports streaming responses, and `@vercel/node` exposes `req.body` via a lazy getter the MCP path never touches, so `getRawBody` reads the intact stream (which is why `oauth-e2e` connected + called `list-pods` against the preview). Residual nit below (L1).

| # | Item | Where | Note / fix |
|---|------|-------|------------|
| L1 | Pass parsed body to `handleRequest` | `src/http.ts:123` | Works today only because nothing reads `req.body` first. SDK's own example is `handleRequest(req, res, req.body)`. Be explicit — cheap insurance against a future refactor silently breaking the transport. |
| L2 | Auth code + handoff URL logged unconditionally | `api/index.ts:216,239` | `oauth_authorize`/`oauth_authorize_handoff` log the `requestId` (= the live auth code) and full handoff URL to prod logs, no DEBUG gate. (Good: the minted key and bearer token are never logged.) Gate behind a verbose flag; drop the handoff URL. |
| L3 | `version=dev` on the hosted path | `src/server.ts`, `src/tools.ts` | `__PACKAGE_VERSION__` is replaced by tsup for the npm/stdio build, but Vercel compiles `api/index.ts → src/*` itself so the define never runs → hosted server emits `User-Agent: runpod-mcp-server/dev`. Observability only. Read version from package.json at runtime in `api/index.ts`. |
| L4 | `stream-job` swallows real error as "timed out" | `src/tools.ts` poll loop | If `serverlessRequest` errors non-consecutively until the 5-min cap, it returns `pollingTimedOut:true` with the last *successful* result — the real error (e.g. 404 job-gone) is discarded. Loop termination itself is correct. Retain `lastError` and include it in the timeout payload. |
| L5 | CORS headers double-set | `vercel.json` headers + `api/index.ts` `res.setHeader` | Both set `Access-Control-Allow-Origin: *`; Vercel applies config headers in addition to function-set ones → possible duplicate `Allow-Origin` (browsers reject `*, *`). Harmless for Node MCP clients, but could break a browser-based remote connector. Pick one source of truth (also: handler sets `Expose-Headers`, vercel.json doesn't). |
| L6 | `encodeBody` mishandles a `Buffer` body | `api/index.ts:115` | A `Buffer` is `typeof 'object'` → `Object.entries` yields byte-index junk → `grant_type`/`code` null → 400. Edge-only (Vercel parses form/JSON into objects). Add `if (Buffer.isBuffer(body)) return body.toString();`. |
| L7 | Per-request `server`/`transport` never disposed | `src/http.ts:111-123` | No `close()` / `res.on('close')` cleanup; on a warm Vercel instance transports accumulate listeners (possible `MaxListenersExceededWarning` / slow memory growth). Dispose on response close. |
| L8 | `getBaseUrl` → `https://undefined` if `Host` missing | `api/index.ts:6-13` | Unlike `src/http.ts:getBaseUrl` (which throws), this doesn't guard a missing host — used for routing + discovery URLs, so it'd silently emit `authorization_endpoint: https://undefined/authorize`. Mirror the http.ts guard. |
| L9 | Tools' GraphQL host not env-configurable | `src/tools.ts:17` | REST + Serverless hosts became env-overridable, but the GraphQL host used by `list-gpu-types`/`list-data-centers` stays hardcoded to prod — a dev deployment hits prod GraphQL even when REST points at dev. (Separate from `RUNPOD_GRAPHQL_URL` in `api/index.ts`, which only affects the OAuth flash flow — two different GraphQL configs.) |
| L10 | `outputDirectory: "."` | `vercel.json` | Ships the whole repo as static assets. Wasteful, not a security hole for a public repo. Add `.vercelignore` or point at an empty dir. |
| L11 | CI pnpm version pin | `.github/workflows/*.yml` | `pnpm/action-setup version: 10` with no `packageManager`/`engines.pnpm` — works today; pnpm 10→11 could shift lockfile expectations and there's no pinned version for local/CI parity. |
| L12 | `oauth-e2e.ts` has no runnable npm script + README pointer wrong | `package.json`, `README.md` | README says "verify with `scripts/oauth-e2e.ts` (see pnpm-free usage in the file header)" but there's no `oauth-e2e` script and the header documents no run command. Add `MCP_SERVER_URL=… npx tsx scripts/oauth-e2e.ts`. |
| L13 | Docs misattribute OAuth routes to `src/http.ts` | `README.md`, `CLAUDE.md` | All OAuth logic lives in `api/index.ts`; `src/http.ts` only does bearer extraction + MCP session. Describe `api/index.ts` as "Vercel adapter + OAuth authorization-server routes." |
| L14 | Contributor docs miss the new env vars | `docs/context.md`, `CLAUDE.md` | All 6 new env vars are in `README.md` but absent from the contributor "engineering reference" docs. |
| L15 | `oauth-e2e.ts` double-decodes the handoff `redirect` param | `scripts/oauth-e2e.ts:73-78` | `searchParams.get('redirect')` already decodes once, then `decodeURIComponent` again. Works today only because the inner value has no surviving `%`-sequences. Debug-harness only. |

**FYI (not a bug):** the GraphQL string interpolation in `api/index.ts` (`createFlashAuthRequest`, `getFlashAuthStatus`) is injection-safe via `JSON.stringify` of string literals — using GraphQL variables would be the robust idiom but isn't required.

**Verified good:** all 34 tools identical old→new (names/schemas/endpoints/methods/bodies); auth threading correct; env-var defaults match old hardcoded hosts exactly; npm publish path + `npx` install + bin all intact (`npm pack` checked); changesets — nothing dropped (the 2 deleted ones were already in the 1.2.0 changelog); no stale `RUNPOD_HTTP_SHARED_API_KEY`/`jose`/Clerk refs.

TimPietruskyRunPod:
@justinwlin — addressed all the review feedback; ready for another pass.

**High**
- `redirect_uri` is now validated against an allowlist (registered client callbacks + loopback) before the authorization code is ever issued.

**Mediums (m1–m8)**
- m1: dropped the PKCE (`S256`) advertisement — it was never enforced server-side. Verified live that a mismatched `code_verifier` still mints, confirming it was inert.
- m2: confirmed against the backend (`model/src/flash/authRequests.ts`) that a guest read of `flashAuthRequestStatus` atomically flips `APPROVED→CONSUMED` and returns the key exactly once; documented in `/token`.
- m3: removed `MCP_OAUTH_ENABLED` — the hosted server always advertises the OAuth challenge; bring-your-own-key callers are unaffected. Verified local stdio still works with just `RUNPOD_API_KEY`.
- m4: minted key name now defaults to `runpod-mcp` (`RUNPOD_API_KEY_NAME`, `""` to suppress). Scoped/expiring keys remain a backend follow-up.
- m5: `CONSOLE_BASE_URL` defaults to `https://console.runpod.io`.
- m6: caller-tracking falls back to the inbound HTTP `User-Agent` on stateless HTTP (no more `client=unknown`).
- m7: `/token` polls ~45s instead of a non-retryable `authorization_pending`.
- m8: `.changeset/` is no longer gitignored (it's the release source).

**Lows (L1–L15)** — all done: explicit body pass + per-request transport disposal, `Buffer` body handling, `Host` guard, runtime package version in both the User-Agent and McpServer metadata, `RUNPOD_PUBLIC_GRAPHQL_URL`, single CORS source, empty static output dir, `engines.pnpm`, `oauth-e2e` script, doc/route-attribution fixes, and the harness double-decode.

**Verification:** CI green on Node 18/20/22; full OAuth chain (discovery → DCR → authorize → console approval → redirect-back → `/token` → minted `runpod-mcp` key) exercised end-to-end against a preview deploy; stdio + HTTP smoke pass.

Only deliberate non-fixes are the backend-side features (scoped/expiring keys), noted as follow-ups.

justinwlin (approved):
Reviewed updated PR — all HIGH and MEDIUM findings from prior review are addressed in code:

- redirect_uri now exact-match allowlisted at both /authorize and /token (loopback per RFC 8252) — closes the auth-code/API-key theft vector
- PKCE no longer advertised (not enforced server-side)
- MCP_OAUTH_ENABLED flag removed; allowlist makes always-live routes safe
- CONSOLE_BASE_URL defaults to production console
- minted key defaults to name 'runpod-mcp' (identifiable/revocable)
- auth code only logged behind MCP_VERBOSE_LOGS
- hosted version read from package.json at runtime
- tools GraphQL host env-configurable; vercel.json outputDirectory fixed

CI green (3/3), 34 tools preserved.

Residual tradeoffs (accept, not blockers): access token is a long-lived full-privilege Runpod API key (no expiry/refresh/revocation endpoint), and code replay protection relies entirely on the backend single-use (APPROVED→CONSUMED atomic read). Worth a line in release notes that users revoke the minted key via dashboard.

Approving.

PR #32, opened 2026-06-23
feat: cap list-tool responses + add test suite

lukepiette:
## Summary
Two things, both fixing structural gaps flagged in the Agents Emergency Triage PRD:

1. **Stop list tools from blowing up the agent's context.** The MCP `list-*` tools serialized the full REST response, and the REST list endpoints have no server-side pagination, so a large account returns a payload many times an LLM's context window (`list-templates` observed at ~15x Claude Code's window), ending the session. This adds a **client-side cap**.
2. **Bootstrap a test suite** (the repo had none) so this logic — and future generated tools — are gated in CI.

### Pagination / token caps
- New `src/pagination.ts` module: `capListResult` + shared `listPaginationParams` (optional `limit`/`cursor`).
- Applied to all five list tools: `list-pods`, `list-endpoints`, `list-templates`, `list-network-volumes`, `list-container-registry-auths`.
- Responses return at most `limit` items (default 20, max 100) in a `{ items, pagination }` envelope reporting `total`, `returned`, `offset`, `truncated`, and a `nextCursor`.
- Optional params, so existing callers that pass nothing keep working (default page).

### Test suite (new)
- Node's built-in test runner via `tsx` — no new dependency. Added `pnpm test`.
- `tests/pagination.test.ts`: cursor round-trip + garbage/negative handling, default/explicit limit, clamp to ceiling, multi-page walk, empty list, non-array passthrough, cursor past end.
- `tests/tools-registration.test.ts`: drives the real `registerTools`; asserts the core tool set registers, no duplicate tool names, and every `list-*` tool exposes `limit`/`cursor`.
- `scripts/smoke-list.ts`: live end-to-end check (boots stdio server, real MCP client, calls each list tool) for manual verification.

## Forward-compatibility
REST will add cursor-based pagination (needs upstream GraphQL pagination per endpoint). The `limit`/`cursor` signature here matches it, so the tool interface won't change when server-side pagination lands; the MCP will forward the params and surface a real cursor. The client-side cap stays as a backstop.

## Test plan
- [x] `pnpm test` — 16 tests pass
- [x] `tsc --noEmit` clean
- [x] `tsup` build succeeds
- [x] `eslint` clean
- [x] **Live end-to-end against a real account** via `scripts/smoke-list.ts`: `list-templates` returned 26 total, capped to 20, `truncated: true`, `nextCursor: "MjA="`; paging with that cursor returned the remaining 6 (`offset: 20`, `truncated: false`). The other four list tools were under the cap and returned fully (no over-truncation).

### DIFF
--- a/.changeset/list-tools-pagination-token-caps.md
+++ b/.changeset/list-tools-pagination-token-caps.md
@@ -1,0 +1,15 @@
+---
+"@runpod/mcp-server": minor
+---
+
+Cap list-tool responses so they no longer overflow the agent's context window.
+Every `list-*` tool (`list-pods`, `list-endpoints`, `list-templates`,
+`list-network-volumes`, `list-container-registry-auths`) now accepts optional
+`limit` and `cursor` parameters and returns at most `limit` items (default 20,
+max 100) inside a `{ items, pagination }` envelope that reports `total`,
+`returned`, `truncated`, and a `nextCursor` for fetching the next page.
+
+The REST API does not yet support server-side pagination, so this is a
+client-side cap; the `limit`/`cursor` signature is shaped to match the
+cursor-based pagination the REST API will add, so the tool interface will not
+change when server-side pagination lands.
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "clean": "rm -rf dist *.tsbuildinfo",
     "lint": "eslint \"./**/*.ts*\"",
     "type-check": "tsc --noEmit -p tsconfig.json",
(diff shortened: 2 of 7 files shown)


justinwlin (approved):
Reviewed locally: `pnpm test` 16/16, `tsc --noEmit` clean, `tsup` build success. Logic verified — filters applied server-side via query params, cap applied after (no over-truncation). Clean isolated `pagination.ts`, defensive cursor handling, backward + forward compatible. LGTM.

PR #34, opened 2026-06-23
[Phase A · PR1] Adapter core: unwrapList, resolveVersion, probe, resolveBackend (v1)

justinwlin:
**Stacked draft — do not merge.** Base = \`DO-NOT-MERGE-TILL-V2-LIVE/v2-rest-support\` (drafting workspace). Implements **Phase A / Steps A0–A2** of PLAN.md.

## What this is
The pure, additive routing core for the v1/v2 adapter. **No wiring into tools yet** — `tools.ts` is untouched, so v1 behavior is unchanged. This PR just adds `src/_shared/backend.ts` + tests.

## Contents
- **A0 `unwrapList(resource, version, raw)`** — v1 bare-array passthrough; v2 envelope unwrap (`{pods:[…]}` → array); non-array/missing-key → `[]`. Kills the envelope bug at the seam before v2 lands.
- **A1 `resolveVersion`** — per-resource `RUNPOD_REST_VERSION_<RES>` override → global `RUNPOD_REST_VERSION` → default `v1`. `auto` is stdio-only (http → v1, no cross-user probe leak). `jobs`/`endpoints` pinned v1.
- **A1b `createV2Prober`** — memoized probe, injected `fetch` (offline-testable).
- **A2 `resolveBackend`** — v1 REST descriptors with the **exact** current paths (`/pods`, `/networkvolumes`, `/containerregistryauth`, `/endpoints`), GraphQL `kind` for catalog, serverless base for `jobs`; v2 **throws** until Phase B (keeps Phase A honest).

## Tests / checks
- 25 new offline tests (no network, no SDK); **41 total pass**.
- `tsc --noEmit` clean · `eslint` clean · `tsup` build clean · new files prettier-clean.

## Notes
- `endpoints` correctly uses the **REST v1** base, not the serverless base.
- Pre-existing prettier warning in `tests/tools-registration.test.ts` is on the base branch, left untouched to keep this PR scoped.

### DIFF
--- a/src/_shared/backend.ts
+++ b/src/_shared/backend.ts
@@ -134,39 +134,61 @@ export function resolveVersion(opts: {
 // response) pins true/false, but a transient failure (thrown error or 5xx) is
 // NOT cached — it returns false for this call and re-probes next time, so a
 // one-time startup network blip doesn't pin the process to v1 forever.
-type FetchLike = (
+// Exported so the stdio entrypoint can type its node-fetch adapter against the
+// exact shape the prober needs (just `{ status }`), instead of casting.
+export type FetchLike = (
   url: string,
   init?: { method?: string; headers?: Record<string, string> }
 ) => Promise<{ status: number }>;
 
-export function createV2Prober(deps: {
+export interface ProbeDeps {
   fetch: FetchLike;
   baseUrl: string;
   apiKey: string;
-}): () => Promise<boolean> {
+}
+
+// Classify a probe response status into a definitive verdict, or `undefined` for
+// a transient failure (5xx) that must NOT be cached.
+function classifyProbeStatus(status: number): boolean | undefined {
+  if (status >= 500) return undefined;
+  return status >= 200 && status < 300;
+}
+
+// One probe attempt. Returns a definitive verdict, or `undefined` when the
+// attempt was transient (network throw / 5xx) and should be retried.
+async function runV2Probe(deps: ProbeDeps): Promise<boolean | undefined> {
+  try {
+    const res = await deps.fetch(`${deps.baseUrl}/catalog/gpus`, {
+      method: 'GET',
+      headers: { Authorization: `Bearer ${deps.apiKey}` },
+    });
+    return classifyProbeStatus(res.status);
+  } catch {
+    return undefined;
+  }
+}
+
+// Memoizing v2 prober: probes once, caches only a *definitive* verdict, and
+// coalesces concurrent first-calls onto a single in-flight probe. A transient
+// attempt resolves to false for this call but leaves the cache unset, so the
+// next call re-probes (a one-time startup blip can't pin the process to v1).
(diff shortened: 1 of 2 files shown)


justinwlin:

Two independent reviews (correctness/design + adversarial test-quality). All findings addressed in `6b5b832`:

**[BLOCKER] async-probe type mismatch** — `createV2Prober` returns `() => Promise<boolean>` but `resolveVersion` took `probe: () => boolean`; a Promise is always truthy, so wiring the real prober would force `auto`→v2 unconditionally. **Fix:** `resolveVersion` now takes a resolved `v2Available?: boolean` (no thunk). The startup wiring awaits `createV2Prober()` once (stdio only) and passes the boolean. Footgun gone.

**[FIX] transient probe failures cached permanently** — a one-time startup blip pinned the process to v1 forever. **Fix:** only *definitive* verdicts (non-5xx responses) are cached; 5xx/thrown errors return false but re-probe next call. Added recovery tests. Also coalesces concurrent first-calls onto one in-flight probe.

**[FIX/NIT] coverage gaps** — added: false(401)-memoization (the `cached !== undefined` guard), transient-not-cached recovery, probe URL+bearer-auth assertion, 2xx/<300 bounds, per-resource v1-over-v2, case-insensitivity (`V2`/`AUTO`), bogus-value fallback, auto+stdio with no verdict, `get(id)` for every REST resource, serverless/graphql base env-overrides, jobs/catalog expose no list/get path, every-resource-resolves guard, `undefined` raw, identity reference-equality.

**Confirmed correct (no change):** v1 descriptor paths/bases match `tools.ts` exactly (cited); all 7 v2 envelope keys match the spec; `endpoints` on REST base not serverless; `resolveBackend` throwing on v2 is the right Phase-A guard.

Tests: **16 → 56** (40 new offline). `tsc`/`eslint`/`tsup` clean. Running a final verification pass.

justinwlin:

Independent verification subagent confirmed: both review fixes correct and complete, no new race/leak, no incorrect-v2 path, additive-only (zero importers — v1 untouched), v1 descriptor paths spot-checked against `tools.ts` and match. `pnpm test` 56 pass · `tsc`/`eslint` clean.

Ready for review. Next stacked PR (A3 unified HTTP client) builds on this branch.

justinwlin:
**Cleanup relocated here (no behavior change).** The `createV2Prober` cleanup now lives in this PR, where the function is introduced — replaced the async-IIFE-in-`finally` with two pure helpers (`classifyProbeStatus` + `runV2Probe` returning `boolean | undefined`) and a 4-line memoizer; added `ProbeDeps` and **exported `FetchLike`**. (The startup timeout is still added in the C3 PR, slotted into `runV2Probe`.) Same caching/coalescing semantics; tests green.

justinwlin:
Superseded by #44 (v2 1/3 — Adapter foundation), which bundles this into one reviewable unit. Branch retained.

PR #35, opened 2026-06-23
[Phase A · PR2] Unified HTTP client + pure tracking headers (A3)

justinwlin:
**Stacked draft — do not merge.** Base = `v2/phase-a1-adapter-core` (PR #34). Implements **Phase A / Step A3**.

## What this is
The unified authenticated JSON client that replaces the byte-identical `runpodRequest` + `serverlessRequest`, plus a pure home for tracking headers. **Additive** — not wired into `tools.ts` yet (A4/PR3 does that), so v1 behavior is unchanged.

## Contents
- **`src/_shared/http.ts` — `createHttpClient({apiKey, fetch, tracking, errorPrefix})`**: takes a fully-resolved URL (the adapter builds base+path, collapsing the old `endpointId`-split), injected `fetch` so it's offline-testable.
- **`HttpError(status, body)`** — thrown on `!ok`. **Refines the PLAN's provisional 501 contract:** instead of *returning* `{unsupported}` (which would defang `stream-job`'s consecutive-error counter — the A4 risk), it **throws** a typed error carrying `.status`, so `create-pod` (C2) branches on 501 while `stream-job` still counts it. Best of both.
- Content-type matches the **`+json` suffix** so v2's RFC-9457 `application/problem+json` errors are parsed, not swallowed; `204`/empty/non-JSON → `{success:true,status}`.
- **GraphQL intentionally NOT merged** (no auth header, own `{data,errors}` envelope) — stays separate through Phase A.
- **`src/_shared/tracking.ts`** — pure `sanitizeUaToken` + `buildTrackingHeaders` (SDK-free; the home A4 wires the handlers into).

## Tests / checks
- 14 new offline tests (injected fetch, no network/SDK): request shape (URL/headers/method/body), response handling (json, problem+json, 204/empty, HttpError w/ status+body, 501, per-client prefix), tracking header construction + fallback.
- **70 total pass** · `tsc`/`eslint`/`tsup` clean.

## Deviation note (intentional)
The 501 handling deviates from PLAN A3's "returns {unsupported:true}" — see HttpError rationale above. This is the explicitly-"provisional" contract being resolved in favor of the design that doesn't break stream-job. PLAN can be updated to match.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -311,8 +311,9 @@ GREEN, or their RED tests have nothing injectable to target. This is stated per-
   takes `fetch` as a param (today it's a module import, `tools.ts:3`); thread from `stdio.ts`/`http.ts`.
 - RED: `tests/http.test.ts`. Inject a fake `fetch`. Cases: builds `base+path`; sets
   `Authorization: Bearer <key>` + tracking headers; serializes body only for POST/PATCH;
-  `!ok → throws` with the configured error prefix (**provisional contract — C2 carves out
-  501 as the one non-throwing case; note that here so C2 amends rather than regresses**);
+  `!ok → throws HttpError` (carrying `.status` + `.body`) for **all** non-OK responses
+  including 501 — C2 *catches* the 501 throw and branches on `err.status`, so stream-job's
+  error counter still fires on a mid-poll 501;
   non-JSON / 204 / empty body → `{success:true,status}`; JSON → parsed; **`application/
   problem+json` (v2's error shape) parses as JSON** — match the `+json` *suffix*, NOT the
   literal substring `application/json` (`"application/problem+json".includes("application/json")`
--- a/PLAN.md
+++ b/PLAN.md
@@ -299,11 +299,14 @@ GREEN, or their RED tests have nothing injectable to target. This is stated per-
 
 **Step A3 — Unified HTTP client, injected `fetch`. (SCOPE: REST + Serverless only — NOT GraphQL.)**
 - **CONTRACT (pin this first — it threads through A3→A4→B4→C2):** the client is **low-level**,
-  signature `client(path, method?='GET', body?): Promise<unknown>` (mirrors today's
-  `runpodRequest`, `tools.ts:126-130`). On `!ok` it **throws** (stream-job's loop depends on
-  it, A4) — **except 501**, which it returns as `{unsupported:true,status:501}` (C2). It takes
-  a per-client `errorPrefix` option so `"Runpod API Error"` / `"Runpod Serverless API Error"`
-  / `"GraphQL Error:"` are preserved (decide here, not at A5).
+  signature `client(url, method?='GET', body?): Promise<unknown>` (takes a fully-resolved URL;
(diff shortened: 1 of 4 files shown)


justinwlin:

**[BLOCKER] real `!ok`+problem+json error path was untested** — the old test used an impossible `200`+problem+json. **Fixed:** added a real `400`+problem+json test asserting the body is read via `.text()`, `.json()` is **never** called, and the JSON text lands on `HttpError.body`; reframed the success test to a `application/vnd.api+json` vendor type.

**[BLOCKER] `tracking()` per-request invocation untested** — a memoization refactor could break per-call session ids silently. **Fixed:** added a test asserting tracking is re-evaluated every request (session id differs per call).

**[BLOCKER→doc] 501 strategy vs PLAN** — reviewer confirmed the code is **zero-behavior-change** (today every `!ok` throws a generic Error; throwing `HttpError` on 501 is byte-compatible). The mismatch was with PLAN's *provisional* `{unsupported}` return, which would have defanged stream-job's error counter (the A4 risk). **Resolved by reconciling PLAN A3+C2** to the implemented design: client throws `HttpError`; create-pod catches and branches on `err.status===501`. `HttpError.status` exists precisely for this.

**[FIX/NIT] edge coverage added:** charset suffix + case-insensitive content-type, `+json` vendor type, PUT-no-body, `HttpError.name`, sanitize 64/65 boundary + all-reserved/empty, empty-string clientName → `unknown` (`||` not `??`).

**Confirmed by review (no change):** headers/body-gating/content-type branch line-for-line equivalent to old helpers; `isJsonContentType` is a strict superset of the old check (fixes the latent problem+json swallow); `buildTrackingHeaders` reproduces the UA string char-for-char; node-fetch `Response` satisfies the interfaces (A4 wiring compiles). A4 dependencies noted (URL resolution, two clients w/ distinct prefixes, caller wires tracking).

Tests: **70 → 76**. `tsc`/`eslint`/`tsup` clean. Running final verification.

justinwlin:

Verification subagent: `pnpm test` **76 pass**, `tsc`/`eslint` clean. Confirmed: error path reads `.text()` and never `.json()`; `isJsonContentType` matches all JSON/+json/charset/case variants and rejects text-plain/null; body only serialized for POST/PATCH; per-request tracking test genuinely proves non-memoization; module still additive (zero runtime importers). Fixed the one leftover stale 501 sentence in PLAN A3. No blockers remain.

justinwlin:
**Cleanup pass (code-quality refactor, no behavior change).** Tightened `createHttpClient` internals per the a/b/c review (remove nesting, stricter types, helpers):

- Named the request-init shape as `RequestInitLike` and reused it for `FetchLike`'s param, so the inline `init` literal can no longer drift from what `fetch` accepts (stricter types).
- Extracted `methodSendsBody()` and `buildRequestHeaders()` helpers; flattened the JSON-vs-success-marker branch into one expression.

Same outbound requests, same `HttpError`, same success marker. Verified by the existing http + handler tests (167 on the tip after the stack rebase) plus an independent reviewer pass — no drift.

justinwlin:
Superseded by #44 (v2 1/3 — Adapter foundation), which bundles this into one reviewable unit. Branch retained.

PR #37, opened 2026-06-23
[Phase B · PR4] v2 request mappers + v2 resolveBackend descriptors (B1/B2/B3)

justinwlin:
**Stacked draft — do not merge.** Base = `v2/phase-a4-wire-adapter` (PR #36). Implements **Phase B Steps B1, B2, B3** as a pure, fixture-tested v2 translation layer. Handler wiring (routing through `resolveBackend`) and catalog REST response mapping (B5) are follow-ups.

## Contents
- **`src/_shared/mappers.ts`** — pure v1→v2 request-body mappers:
  - `mapPodCreateToV2`: ContainerConfig flatten (`imageName→image`, `containerDiskInGb→disk`, `volumeInGb/volumeMountPath→mounts.persistent.{size,path}`), `gpuTypeIds[0]→gpu.id` + `gpuCount→gpu.count` (array→scalar), `cloudType→cloud`, `dataCenterIds[0]→dataCenter`, **never emits both `gpu`+`cpu`**, drops unknown/undefined keys.
  - `mapPodUpdateToV2`, `mapTemplateCreateToV2` (`isServerless→serverless`), `mapNetworkVolumeCreateToV2` (`dataCenterId→dataCenter`), registry identity.
- **`resolveBackend` v2 branch**: real v2 paths (`/v2/pods`, `/v2/network-volumes` **hyphen**, `/v2/registries`, `/v2/catalog/*` incl. **non-hyphenated** `datacenters`), version-aware `unwrap`, per-resource `mapCreate`/`mapUpdate`. `jobs`/`endpoints` still pinned v1.
- **`tests/fixtures/v2-create-pod.json`** — committed golden for the full pod-create remap.

## Spec volatility (documented in the mapper header)
Targets the current v2 dev snapshot. Known pending upstream changes that require updating mapper **+ fixture together**: `dataCenter` scalar may revert to `dataCenterIds[]`; `cloud` may drop `ALL`; template `category` may become optional. Per PLAN Part 7/8.

## Tests / checks
- New: `tests/mappers.test.ts` (fixture + field-level cases) + v2 descriptor tests in `backend.test.ts`. **112 pass.** `tsc`/`eslint`/`tsup` clean.

## Scope notes
- This is the v2 **library**; nothing routes to v2 at runtime yet (handlers still use inline v1 paths). The handler routing through `resolveBackend` is the next PR — that's where v2 actually turns on behind the flag.
[...]

PR #38, opened 2026-06-23
[Phase B · PR5] Route pod handlers through resolveBackend + B4 actions (v2 keystone)

justinwlin:
**Stacked draft — do not merge.** Base = `v2/phase-b1-mappers-descriptors` (PR #37). The **keystone** — pods actually route to v2 behind the flag. Implements **B4** + the handler-routing pattern.

## What changed
- **Descriptor convention bug fixed** (surfaced by wiring): v2 base ends in `/v2` and v2 paths were *also* `/v2/pods` → `base+path` doubled to `/v2/v2/pods`. v2 paths are now **bare** (`/pods`, `/catalog/gpus`, …) so `base+path` = exactly one `/v2`. Matches what the prober already assumed. Added a combined-URL regression test.
- **`registerTools` gains** `deps.v2Available` + `callRestUrl` (full-URL REST client), `backendFor()`, `jsonReply()`, and `podAction()` (B4).
- **All 7 pod handlers routed through `resolveBackend`:**
  - **v1: byte-identical** — filters, paths, bodies, query encoding all preserved.
  - **v2:** `/v2/*` paths, `mapCreate`/`mapUpdate` bodies, `unwrap`+cap lists, and **`POST /v2/pods/{id}/action {action}`** for start/stop (B4). v2 drops the v1-only list/get filter query params (spec declares none).

## Golden coverage (both versions)
- v1 (unchanged): list (+filters), get, create (byte-identical body), update (id-strip), start/stop (`/start`,`/stop`), delete.
- v2: list (`/v2/pods`, no filters, envelope unwrap + cap), create (mapped body — `image`/`gpu`/`dataCenter`), update (mapped, no podId), **start/stop → `/action` with `{action}`**, delete. Asserts the single-`/v2` combined URL.

## Checks
**127 tests pass** · `tsc`/`eslint`/`tsup` clean.

## Scope
Pods only. templates/volumes/registries routing + catalog B5 (GraphQL→REST) + C1/C2/C3 follow in subsequent stacked PRs, using this exact pattern. The stdio entrypoint awaiting `createV2Prober()` to supply `deps.v2Available` for `auto` mode is a small follow-up (explicit `RUNPOD_REST_VERSION=v2` works today).

### DIFF
--- a/src/_shared/backend.ts
+++ b/src/_shared/backend.ts
@@ -242,22 +242,25 @@ const CATALOG: ReadonlySet<Resource> = new Set<Resource>([
   'dataCenters',
 ]);
 
-// v2 REST paths (the control-plane resources). Note the renames vs v1:
-// network-volumes (hyphen), registries (was containerregistryauth), catalog/*.
-// Catalog `datacenters` is intentionally NOT hyphenated (matches the spec).
+// v2 REST paths (the control-plane resources). Paths are relative to the v2
+// base which ALREADY includes `/v2` (restV2Base → https://v2-rest.runpod.io/v2),
+// so they do NOT repeat the `/v2` prefix — `base + list` = `.../v2/pods`. (The
+// prober relies on the same convention: `base + /catalog/gpus`.)
+// Renames vs v1: network-volumes (hyphen), registries (was containerregistryauth),
+// catalog/*. Catalog `datacenters` is intentionally NOT hyphenated (matches the spec).
 const V2_REST_PATHS: Partial<
   Record<Resource, { list: string; get?: (id: string) => string }>
 > = {
-  pods: { list: '/v2/pods', get: (id) => `/v2/pods/${id}` },
-  templates: { list: '/v2/templates', get: (id) => `/v2/templates/${id}` },
+  pods: { list: '/pods', get: (id) => `/pods/${id}` },
+  templates: { list: '/templates', get: (id) => `/templates/${id}` },
   networkVolumes: {
-    list: '/v2/network-volumes',
-    get: (id) => `/v2/network-volumes/${id}`,
+    list: '/network-volumes',
+    get: (id) => `/network-volumes/${id}`,
   },
-  registries: { list: '/v2/registries', get: (id) => `/v2/registries/${id}` },
-  gpus: { list: '/v2/catalog/gpus', get: (id) => `/v2/catalog/gpus/${id}` },
-  cpus: { list: '/v2/catalog/cpus' },
-  dataCenters: { list: '/v2/catalog/datacenters' },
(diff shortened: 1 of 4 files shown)


justinwlin:

**Reviewer 1 (routing correctness): production-ready, no blockers** — confirmed v1 byte-identical for all 7 pod handlers, v2 URLs/bodies/action endpoint correct vs spec, single-`/v2` combined URLs, prober unaffected by the path-convention fix, callRestUrl preserves logging/HttpError, process.env routing safe for stdio+HTTP.

**Reviewer 2 (test completeness):** found a real test-helper bug + gaps, all fixed:
- **[FIX] `withV2` didn't await** — it restored `RUNPOD_REST_VERSION` before the awaited handler ran (passed only by timing luck). Now `async` + `return await fn()` in try/finally.
- **[BLOCKER] get-pod under v2 was the only un-golden-locked pod op** — added a test asserting `GET /v2/pods/{id}` with **NO** query string (guards the 'v2 drops v1-only filters' claim).
- **[FIX]** create-pod v2 now asserts `mounts.persistent` mapping + a partial-volume **drop** (the riskiest mapper branch, now exercised through the handler).
- **[FIX]** per-resource override (`RUNPOD_REST_VERSION_PODS=v2`) exercised **end-to-end through a handler**, not just the resolveBackend unit.
- **[NIT]** combined-URL double-`/v2` guard extended to `get(id)` + `/action` paths.

Tests: **127 → 130**. `tsc`/`eslint`/`tsup` clean. Noted intentional staged item: `createV2Prober` not yet wired into stdio/http, so `auto` resolves v1 today; explicit `RUNPOD_REST_VERSION=v2` works (probe wiring is a later step). Running final verification.

justinwlin:

Verification subagent: `pnpm test` **130 pass**, `tsc`/`eslint`/`tsup` clean. Confirmed: `withV2` awaits correctly (mutation-tested sensitive); all 7 pod ops golden-locked under **both** v1 (byte-identical) and v2 (single-`/v2` URLs, mapped bodies, `/action` for start/stop); B4 routing correct; non-pod handlers untouched; env-cleanup safe on failure. Added the verifier's suggested `beforeEach` env-isolation guard as defense-in-depth. No blockers.

justinwlin:
**Cleanup added here (no behavior change).** `jsonReply` is introduced in this PR, so it's now defined at module scope and every reply routes through it (replacing the inline `content:[{type:'text'...}]` blocks); `callRestUrl` reuses `withApiErrorLog` from the A4 PR. Identical replies + error logs.

justinwlin:
Superseded by #45 (v2 2/3 — Routing), which bundles this into one reviewable unit. Branch retained.

PR #41, opened 2026-06-23
[Phase B/C · PR8] Catalog GraphQL→REST under v2 + list-cpu-types/get-gpu-type (B5/C1)

justinwlin:
**Stacked draft — do not merge.** Base = `v2/phase-c1-c2-newtools-501` (PR #40). Implements **B5** + the remaining **C1** catalog tools.

## Changes
- **`list-gpu-types` / `list-data-centers`**: v1 GraphQL path **unchanged**; under v2 route to REST `/v2/catalog/gpus` / `/v2/catalog/datacenters`, unwrap the envelope, and re-apply the client-side filters against v2 field names (`memory`/`secure`/`community`/`name`; `region` enum for DCs). `includeUnavailable` is a **documented no-op** on v2 (v2 has no realtime `stockStatus`).
- **NEW `list-cpu-types`** (`/v2/catalog/cpus`) and **`get-gpu-type`** (`/v2/catalog/gpus/{id}`): v2-only; under v1 they return a clean "v2 only" message with no network call.

## Tests
- v2 catalog goldens: list-gpu-types (filters on v2 fields + searchTerm), list-data-centers (region enum), list-cpu-types (v1 message + v2 path), get-gpu-type (v1 message + v2 path).
- **159 pass** · `tsc`/`eslint`/`tsup` clean.
- Note: the v1 catalog GraphQL path is intentionally **not** invoked in tests (it uses the module `fetch`, not the injected client, so it would hit the network). v1 catalog behavior is unchanged from `main`.

## Remaining after this
**C3** (wire `createV2Prober` into stdio for `auto` + flip default when v2 prod live) and **B6** (live CRUD smoke). That completes the plan.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #42, opened 2026-06-23
[Phase C · PR9] Wire createV2Prober into stdio for auto mode (C3)

justinwlin:
**Stacked draft — do not merge.** Base = `v2/phase-b5-catalog-rest` (PR #41). Implements **C3** (the `auto`-mode probe wiring; the deploy-time default flip stays a config/env decision).

## Changes
- **`src/stdio.ts`**: if any `RUNPOD_REST_VERSION*` is `auto` (`wantsAutoProbe`), probe v2 once at startup via `createV2Prober` and pass the resolved `v2Available` boolean into `registerTools`. Explicit `v1`/`v2` skips the probe. Startup wrapped in `async main()`.
- **`src/_shared/backend.ts`**: new pure `wantsAutoProbe(env)` helper (unit-tested).
- `auto` remains **stdio-only** — hosted HTTP pins per-deploy and `resolveVersion` already treats `auto` as v1 under `http` (no cross-user probe leak).

## Tests
- `wantsAutoProbe`: global auto (any case), per-resource auto, false for v1/v2/unset/bogus.
- **165 pass** · `tsc`/`eslint`/`tsup` clean. (`createV2Prober` decision logic was already unit-tested in PR #34.)

## Remaining
**B6** — live CRUD smoke against the dev account with teardown (a manual/CI script, not an offline unit test). That's the last item.

### DIFF
--- a/src/_shared/backend.ts
+++ b/src/_shared/backend.ts
@@ -155,14 +155,23 @@ export function wantsAutoProbe(env: Env): boolean {
 // exact shape the prober needs (just `{ status }`), instead of casting.
 export type FetchLike = (
   url: string,
-  init?: { method?: string; headers?: Record<string, string> }
+  init?: {
+    method?: string;
+    headers?: Record<string, string>;
+    signal?: AbortSignal;
+  }
 ) => Promise<{ status: number }>;
 
 export interface ProbeDeps {
   fetch: FetchLike;
   baseUrl: string;
   apiKey: string;
+  timeoutMs?: number;
 }
+
+// Bound the startup probe so a slow-but-reachable v2 host can't hang stdio
+// startup forever — an abort throws, which runV2Probe degrades to a transient.
+export const PROBE_TIMEOUT_MS = 4000;
 
 // Classify a probe response status into a definitive verdict, or `undefined` for
 // a transient failure (5xx) that must NOT be cached.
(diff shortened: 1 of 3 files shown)


justinwlin:

**Reviewer 1: production-ready** — async main() correct, probe gated to auto, key narrowed, http path provably unaffected, probe failures degrade to v1.

**Reviewer 2: found a real [FIX] (hang risk), resolved:**
- **[FIX] no startup-probe timeout** — a slow-but-reachable v2 host would hang stdio startup forever (`node-fetch` has no default timeout, and the probe runs before `server.connect`). **Fixed:** `createV2Prober` now passes `AbortSignal.timeout(4s)` (configurable); an abort throws → the existing catch degrades to v1. Added a test asserting the signal is passed and an aborted probe → false.
- **[NIT]** added wantsAutoProbe empty-string + startsWith over-match coverage.

Confirmed: network-down already degrades to v1 (errors swallowed), default/non-auto path unchanged (no probe), hosted http untouched. Tests **165 → 167**; `tsc`/`eslint`/`tsup` clean. Production-ready.

justinwlin:
**Cleanup pass (code-quality refactor, no behavior change)** on `src/_shared/backend.ts` + `src/stdio.ts` — this is where the flagged `createV2Prober` lived:

- `createV2Prober`: replaced the async-IIFE-in-`finally` with two pure helpers — `classifyProbeStatus(status)` and `runV2Probe(deps)` returning `boolean | undefined` (`undefined` = transient, don't cache) — leaving a 4-line memoizer. Identical caching / coalescing / timeout semantics.
- Added a reusable `ProbeDeps` interface and **exported `FetchLike`** so `stdio.ts` types its node-fetch adapter against it instead of a `Parameters<…>` double-cast.
- `resolveBackend`: extracted `buildV2Backend`/`buildV1Backend` so the function is a one-line version switch instead of three inline descriptor literals.

An independent reviewer confirmed all five prober invariants + every `Backend` descriptor field are preserved; the `.finally` inFlight reset is safe because `runV2Probe` never rejects. 167 tests pass.

justinwlin:
**Updated:** `createV2Prober`/`resolveBackend` cleanups moved to their introducing PRs (#34/#37). The only refactor remaining here is dropping the `Parameters<…>` double-cast on the probe fetch in `stdio.ts` (now typed against the exported `FetchLike`). The startup timeout fix is merged into the cleaned `runV2Probe`.

justinwlin:
Superseded by #46 (v2 3/3 — Capabilities), which bundles this into one reviewable unit. Branch retained.

PR #36, opened 2026-06-23
[Phase A · PR3] Wire unified client + unwrap into tools; outbound golden; CI test (A4/A5/A6)

justinwlin:
**Stacked draft — do not merge.** Base = `v2/phase-a3-http-client` (PR #35). Implements **Phase A Steps A4 (core), A5, A6**.

## Goal: zero v1 behavior change, while routing through the new seams
This wires the adapter (PR1) + unified client (PR2) into `tools.ts` without changing any observable v1 behavior — proven by an outbound-request golden.

## Changes
- **Unified client**: `runpodRequest`/`serverlessRequest` now delegate to `createHttpClient`, preserving the **exact** URLs, error-message strings, and `console.error` logging. (`HttpError.message` == old `Error.message`, so `stream-job` etc. are unaffected.)
- **Tracking dedupe**: `trackingHeaders` delegates to `buildTrackingHeaders` (identical UA + session id output); the SDK `getClientVersion()` touch stays at the call site.
- **List envelope seam**: the 5 list tools route results through `unwrapList(resource, 'v1', result)` before `capListResult`. v1 = bare array → passthrough (no behavior change), but the seam is ready for Phase B to flip `'v1'` to the resolved version.
- **Test seam**: `registerTools(server, ctx, { fetch })` — optional injected fetch (defaults to node-fetch) so handlers are testable offline.

## A5 — outbound-request golden (`tests/handlers.test.ts`)
Drives the **real** `registerTools` with a capturing fake server + injected fake fetch, and pins what each tool puts **on the wire** (an output-only diff can't see this):
- `list-pods` GET `/v1/pods` no body; filter query-param encoding verbatim (`?computeType=GPU&gpuTypeId=A&gpuTypeId=B&includeMachine=true`)
- `get/stop/delete-pod` URLs+methods; **`create-pod` body byte-identical to params**
- `network-volumes` (`/networkvolumes`), `registries` (`/containerregistryauth`) paths unchanged
- `run-endpoint` → **serverless base** `api.runpod.ai/v2/{id}/run` (not REST)
- list unwrap+cap: 25→20 truncation, explicit limit

## A6 — CI
Added a **Test** step (`pnpm test`) to `.github/workflows/ci.yml` (was type-check/lint/build only — the suite never ran in CI).

## Checks
**87 tests pass** · `tsc`/`eslint`/`tsup` clean.

## Scope note
[...]

Issue #3, opened 2025-08-19
list-templates tool exceeds Claude Code's 25k token limit - returns 386k+ tokens without pagination support


DivineFarm:
### Description
The `list-templates` MCP tool returns responses that significantly exceed Claude Code's maximum token limit of 25,000 tokens, making the tool unusable. The response contains 386,560 tokens, which is over 15x the allowed limit.

<img width="884" height="54" alt="Image" src="https://github.com/user-attachments/assets/3d420067-011f-4416-860e-61af97bc56a9" />

### Error Message
```
Error: MCP tool "list-templates" response (386560 tokens) exceeds maximum allowed tokens (25000). Please use pagination, filtering, or limit parameters to reduce the response size.
```

### Steps to Reproduce
1. Install and configure the MCP server in Claude Code
2. Attempt to use the `list-templates` tool
3. The tool fails immediately with the token limit error

### Expected Behavior
The `list-templates` tool should:
- Return a paginated response that stays within the 25k token limit
- Support optional `limit` and `offset` parameters for pagination
- Provide filtering options to reduce the result set
- Return template metadata by default with an option to fetch full details separately

### Actual Behavior
The tool attempts to return all templates at once (386,560 tokens), causing an immediate failure in Claude Code.

### Environment
- **Claude Code Version**: [Add your version - run `claude --version`]
- **MCP Server Version**: [Add version if available]
- **Operating System**: [Your OS]
- **Context**: Production environment with large template library

### Proposed Solutions
1. **Implement pagination** (Recommended):
   ```javascript
   {
     "limit": 50,  // Default to 50 templates per page
     "offset": 0,  // Starting position
     "include_content": false  // Return metadata only by default
   }
   ```

2. **Add filtering options**:
   ```javascript
   {
     "filter": {
       "category": "construction",
       "search": "keyword",
       "date_range": "last_30_days"
     }
   }
   ```

3. **Separate list and detail endpoints**:
   - `list-templates` - Returns template IDs and metadata only
   - `get-template-details` - Fetches full content for specific template(s)

### Impact
This issue makes the `list-templates` tool completely unusable in Claude Code, blocking our ability to browse and select templates programmatically. 

### Additional Context
- Claude Code has a hard limit of 25,000 tokens per MCP tool response ([[Reference](https://github.com/anthropics/claude-code/issues/4002)](https://github.com/anthropics/claude-code/issues/4002))
[...]

PR #44, opened 2026-06-24
[v2] Migrate Runpod MCP to the v2 REST API (default v2, endpoint CRUD, new tools, spec-parity gate, install wizard)

justinwlin:
**Part 1 of 4 — the v2 REST migration, split into reviewable PRs.**

## This PR: Adapter foundation (no behavior change)
Adds the version-routing core and wires it in. v1 stays **byte-identical** (proven by outbound-request golden tests) — no v2 path is taken yet.

- `src/_shared/backend.ts` — adapter core: `Resource` union, `unwrapList`, `resolveVersion`, memoizing `createV2Prober`, `resolveBackend` → `buildV1Backend`/`buildV2Backend`.
- `src/_shared/http.ts` — one unified authenticated JSON client + `HttpError` (throws on !ok incl 501, problem+json aware).
- `src/_shared/tracking.ts` — pure caller-tracking header builder.
- `src/tools.ts` — handlers delegate to the unified client + tracking; `registerTools(server, ctx, {fetch})` test seam.
- `tests/` — outbound golden + http/backend/tracking units; CI `Test` step.

1. **#44 — Adapter foundation** (no behavior change) ← you are here

|---|---|---|

Do not merge until v2 is live. Supersedes the fine-grained drafts #34–#36.


src/_shared/tracking.ts
```diff
@@ -0,0 +1,38 @@
...
+  sessionId: string;
+}
+
+// Builds the `User-Agent` + `X-Runpod-Session-Id` headers. The clientInfo →
+// fallback → 'unknown' resolution is the caller's job (it owns the SDK handle);
+// this function just sanitizes and formats whatever identity it's given.
+export function buildTrackingHeaders(
+  input: TrackingInput
+): Record<string, string> {
+  const name = sanitizeUaToken(input.clientName || 'unknown');
```
LarveyOfficial, line 31: 🔵 Low: the empty-clientName fallback to the inbound User-Agent is now split across two layers — the call site pre-resolves `info?.name || ctx.clientUserAgent || 'unknown'` and this then re-applies `|| 'unknown'`. Works, but it's fragile: please confirm the HTTP entrypoint passes `clientName` the same way, otherwise the inbound-UA fallback silently degrades to 'unknown' on that transport.

src/_shared/http.ts
```diff
@@ -0,0 +1,121 @@
...
+      headers: buildRequestHeaders(deps.apiKey, deps.tracking),
+    };
+    if (body && methodSendsBody(method)) {
+      init.body = JSON.stringify(body);
+    }
+
+    const response = await deps.fetch(url, init);
+
+    if (!response.ok) {
+      throw new HttpError(
```
LarveyOfficial, line 110: 🔵 Low: a non-JSON 2xx body is silently discarded here (we return only {success, status}). This matches the legacy helpers exactly so it's not a regression — flagging for Phase B in case any v2 endpoint returns a meaningful non-JSON body.

src/_shared/http.ts
```diff
@@ -0,0 +1,121 @@
...
+
+// Thrown on any non-OK response. Carries `status` + `body` so callers can
+// branch (e.g. create-pod maps a 501 to a clean "CPU pods not yet supported"
+// message) without parsing the message string. Because it still THROWS, loops
+// that count consecutive errors (e.g. stream-job) keep working — a 501 mid-poll
+// is surfaced as an error, not swallowed.
+export class HttpError extends Error {
+  readonly status: number;
+  readonly body: string;
+  constructor(prefix: string, status: number, body: string) {
```
LarveyOfficial, line 40: 🔵 Low: the full upstream response body is embedded in error.message and later console.error'd. Same as legacy behavior; flagging only in case an upstream error body ever echoes a token/secret into logs.

src/tools.ts
```diff
@@ -119,111 +113,97 @@ async function graphqlRequest<T>(
...
+// for REST, `endpointId`+path for serverless) and the existing `console.error`
+// log + re-throw behavior. The unified client throws `HttpError` (a subclass of
+// Error) with the SAME message string as before, so callers that read
+// `error.message` (e.g. stream-job) are unaffected.
+
+// The fetch implementation the unified client uses. Defaults to node-fetch;
+// tests inject a fake to capture outbound requests offline (the A4 seam).
+type HttpFetch = Parameters<typeof createHttpClient>[0]['fetch'];
+const defaultFetch = fetch as HttpFetch;
+
```
LarveyOfficial: 🔵 Low/nit: unchecked cast bridging node-fetch's types to the internal FetchLike. Runtime shapes line up today, but a future node-fetch major bump won't be caught by the compiler — consider a thin typed adapter later.

LarveyOfficial (commented):
Automated review pass — read the full diff. Clean, behavior-preserving work with strong golden/wire-lock tests; no blocking issues. A few low-severity notes for later phases are left inline. Verified clean: API key only ever in the Authorization: Bearer header (never logged, never in URLs/query strings); no injection/SSRF/path-traversal (base URLs from a fixed env allowlist, paths from a static map); no floating promises; deliberate and tested `||` vs `??` semantics.

src/_shared/tracking.ts
justinwlin: Confirmed — after the per-resource split this fallback is resolved in ONE place, the shared runtime (`src/tools/runtime.ts:84`): `clientName: info?.name || ctx.clientUserAgent || 'unknown'`. Both transports go through it, and the HTTP entrypoint sets `clientUserAgent` from the inbound User-Agent (`src/http.ts:85`). So the inbound-UA fallback holds on HTTP too — no longer split/fragile.

src/_shared/http.ts
justinwlin: Noted — matches the legacy helper exactly (tested), and no current v2 endpoint returns a meaningful non-JSON 2xx body. Flagged for Phase B if that changes.

src/_shared/http.ts
justinwlin: Noted — same as legacy behavior; Runpod error bodies don't echo tokens. Will keep in mind if an upstream ever does.

src/tools.ts
justinwlin: Noted — fine today; a thin typed adapter is a reasonable future hardening if node-fetch's types shift on a major bump.

src/_shared/backend.ts
```diff
@@ -0,0 +1,295 @@
...
+  pods: 'pods',
+  templates: 'templates',
+  networkVolumes: 'networkVolumes',
+  registries: 'registries',
+  gpus: 'gpus',
+  cpus: 'cpus',
+  dataCenters: 'dataCenters',
+  endpoints: 'endpoints',
+  jobs: 'jobs',
+};
```
jebenexer, line 61: nit: `V2_UNWRAP_KEY` includes entries for `endpoints` and `jobs`, but both are in `V1_ONLY` so they can never reach the v2 unwrap path. Consider dropping those two keys — keeping them implies v2 list-unwrapping is possible for these resources, which contradicts the V1_ONLY invariant a few lines up.

src/_shared/backend.ts
```diff
@@ -0,0 +1,295 @@
...
+// resulting boolean to resolveVersion as `v2Available`. Never wire into the HTTP
+// path (see resolveVersion).
+//
+// Only *definitive* verdicts are cached: a reachable endpoint (any non-5xx
+// response) pins true/false, but a transient failure (thrown error or 5xx) is
+// NOT cached — it returns false for this call and re-probes next time, so a
+// one-time startup network blip doesn't pin the process to v1 forever.
+// Exported so the stdio entrypoint can type its node-fetch adapter against the
+// exact shape the prober needs (just `{ status }`), instead of casting.
+export type FetchLike = (
```
jebenexer: nit: `FetchLike` is also defined in `src/_shared/http.ts` (unexported, but same name). The two shapes are incompatible — this one has all-optional `init` fields; the http.ts one requires `method` and `headers`. Not a runtime issue since they're separate modules, but the naming collision will confuse anyone maintaining code that imports from both. Consider a distinct name here — e.g. `ProbeFetch` — to make the different contracts obvious.

.github/workflows/ci.yml
```diff
@@ -38,6 +38,9 @@ jobs:
       - name: Lint
         run: pnpm run lint
 
+      - name: Test
+        run: pnpm test
```
jebenexer, line 44: Does `pnpm test` require the build output to have been produced first? The test files import from `.js` extensions (`backend.js`, etc.) which suggests either a ts-to-js step or a loader like `tsx`. Running Test before Build is fine if the test runner handles TypeScript directly, but if it depends on `pnpm build` output this order would silently break. Worth a quick confirmation.

justinwlin:
Thanks @jebenexer — dispositions on the three notes:

1. **`V2_UNWRAP_KEY` `endpoints`/`jobs` entries** — keeping them. The map is typed `Record<Resource, string>` (total), so the keys can't be dropped without switching to `Partial`. More importantly, later in the stack (#47) `endpoints` is migrated to v2 (`/v2/serverless`) and **does** use its unwrap key — only `jobs` stays v1-only. So the entries aren't dead in the merged result; no change.

2. **`FetchLike` name collision** (this module's exported probe type vs the stricter unexported one in `_shared/http.ts`) — good catch, fixed: the probe type is renamed **`ProbeFetch`** (the `{ status }`-only contract), with a comment pointing at the distinction. (backend.ts type def + `ProbeDeps.fetch` + the two `stdio.ts` references.)

3. **CI Test-before-Build** — intentional and now documented. `pnpm test` runs via `node --import tsx --test`, which compiles TS on the fly, so it needs no build output. Added a comment in `ci.yml` above the Test step so the ordering isn't mistaken for a latent break.

**Where the fixes landed:** to avoid cascading a 4-branch rebase across the whole stack (#44→#45→#46→#47→#49) for two cosmetic nits, I applied #2 and #3 on the **stack tip (#47, `v2/group-4-source-split`)** — commit `94a92f8` — so they're present in the merged result; #49 was rebased on top. When the stack eventually squashes/merges, #44's own diff is unaffected but the final tree carries the fixes. tsc/eslint/248 tests green.

src/_shared/http.ts
jebenexer, line 1: The `methodSendsBody` check excludes `PUT`. No current v2 endpoint uses PUT, but this client is now the single unified adapter for all future REST calls. A future tool that passes `method='PUT'` with a body will have it silently dropped — the request goes out without a payload and no error is raised. Either add `PUT` here, or add a runtime assertion that callers only pass POST/PATCH when they intend a body.

src/_shared/backend.ts
jebenexer, line 1: A 401 or 403 from the probe is cached as a definitive `false` for the process lifetime — 5xx is correctly left un-cached, but 4xx is treated the same as a confirmed "v2 does not exist" verdict. If the probe URL is transiently gated during a rollout, or the key lacks v2 scope and is later expanded, the process is stuck on v1 with no recovery path short of restart. Consider treating 401/403 as transient (`return undefined`) so the probe can recover, or at minimum log a warning so the permanent pin is visible in logs.

jebenexer (commented):
Follow-up review of the full #44–#49 stack. Two items here that are actionable now — neither is a regression from the legacy helpers, but both should be addressed before the stack merges since this client becomes the single unified adapter for all future v2 calls.

src/_shared/backend.ts
justinwlin: Keeping these — the map is typed `Record<Resource, string>` (total), so the keys can't be dropped without switching to `Partial`. And on the stack tip (#47) `endpoints` is migrated to v2 (`/v2/serverless`) and **does** use its unwrap key; only `jobs` is truly v1-only. So they're not dead in the merged result. No change.

src/_shared/backend.ts
justinwlin: Fixed in `94a92f8` — renamed the exported probe type to **`ProbeFetch`** (the `{ status }`-only contract), with a comment pointing at the distinction from http.ts's stricter `FetchLike`. Updated the type def, `ProbeDeps.fetch`, and the two `stdio.ts` references.

.github/workflows/ci.yml
justinwlin: Confirmed it doesn't need the build: `pnpm test` = `node --import tsx --test`, which compiles TS on the fly (the `.js` import specifiers resolve to `.ts` source). Added a comment above the Test step in `ci.yml` (`94a92f8`) so the order isn't mistaken for a latent break.

src/_shared/http.ts
justinwlin: Fixed in `29361d0` — `methodSendsBody` now includes `PUT`, so a future PUT-with-body isn't silently dropped. (Still no PUT endpoint today, but this is now the single unified client.)

src/_shared/backend.ts
justinwlin: Fixed in `29361d0` — `classifyProbeStatus` now treats **401/403 as transient** (returns `undefined`, not cached), alongside 5xx. So a key whose v2 scope is granted later, or transient gating during a rollout, recovers on the next probe instead of pinning the process to v1 until restart. Added a test asserting 401 → re-probes → recovers to `true`.

jebenexer (approved):
Both items from the follow-up review verified fixed in 29361d0: `methodSendsBody` now includes PUT, and `classifyProbeStatus` treats 401/403 as transient so a key whose v2 scope is later expanded can recover without a restart. Clean.

PR #43, opened 2026-06-23
[Phase B · PR10] Live CRUD smoke with teardown (B6)

justinwlin:
**Stacked draft — do not merge.** Base = `v2/phase-c3-probe-wiring` (PR #42). Implements **B6** — the final plan item.

## What this is
`scripts/smoke-crud.ts` + `pnpm smoke:crud` — a **live** validation that boots the stdio MCP server once per requested REST version and runs a real create→get→delete lifecycle against the dev account, with teardown baked in. Not part of `pnpm test` (it hits the live API and needs `RUNPOD_API_KEY`).

## Safety / teardown (the plan's B6 contract)
- **Free resources only** — templates + container-registry-auth. No pods or network volumes, so it can't incur GPU/storage cost.
- **Unique run prefix** `mcp-smoke-<uuid>` — identifiable, collision-free.
- **Pre-sweep** — deletes leftover `mcp-smoke-*` from a prior crashed run before starting.
- **`try/finally` teardown** — every created id is deleted in `finally`, best-effort (one failure never strands the rest).
- **Fail-closed post-verify** — lists remaining resources; if any run-prefixed resource survives, throws → exit non-zero (a leak fails the gate).
- Loops `v1`/`v2` (`pnpm smoke:crud v1 v2`); v2 needs `RUNPOD_REST_V2_API_URL=https://v2-rest.runpod.dev/v2`.

## Checks
`tsc`/`eslint`/`tsup` clean; **167 unit tests still pass** (script is excluded from the offline suite).

## This completes the PLAN
All work items A0–C3 + B5/B6 are now implemented across PRs #34–#43.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #39, opened 2026-06-23
[Phase B · PR6] Route template/volume/registry handlers through resolveBackend

justinwlin:
**Stacked draft — do not merge.** Base = `v2/phase-b4-pods-routing` (PR #38). Applies the proven pod-routing pattern to the remaining control-plane CRUD resources (14 handlers).

## Changes
All template / network-volume / registry handlers now route through `resolveBackend` + `callRestUrl`:
- **v1: byte-identical** — paths (`/templates`, `/networkvolumes`, `/containerregistryauth`), bodies, id-strip, and the template `include*` query all preserved.
- **v2:** `/v2/templates`, `/v2/network-volumes` (hyphen), `/v2/registries` (rename); `mapCreate`/`mapUpdate` bodies (`serverless`+`category:NVIDIA`, `dataCenterId→dataCenter`, registry identity); list responses `unwrap`+capped; v1-only template filters dropped on v2.
- **Unchanged:** endpoints CRUD + serverless job tools stay on `runpodRequest`/`serverlessRequest` (v1-pinned — no v2 home).

## Golden coverage
- v1 (pre-existing, still green): create-template, create/update/list network-volume, list registries.
- v2 (new): list-templates (no include* query), create-template (mapped body), list/create/get network-volume (hyphen + dataCenter rename), list/create/delete registries (rename + identity).

## Checks
**138 tests pass** · `tsc`/`eslint`/`tsup` clean.

## Remaining after this
B5 (catalog GraphQL→REST + response mappers), C1 (restart / list-cpu-types / get-gpu-type), C2 (501→clean message), C3 (probe wiring + flip default), B6 (live CRUD smoke).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #45, opened 2026-06-24
[v2 2/4] Route every resource through the adapter (v1 + v2)

justinwlin:
**Part 2 of 4 — the v2 REST migration.**

## This PR: Route every resource through the adapter (v1 + v2)
Makes each resource version-aware. v1 paths/bodies unchanged; v2 paths + body mappers added behind the adapter, and every tool gains a description + MCP annotations.

- `src/_shared/mappers.ts` — v1→v2 request-body mappers (pod / template / network-volume) + `persistentMount`/`gpuConfigToV2` helpers.
- v2 descriptors in `backend.ts` (`V2_REST_PATHS`, `V2_MAPPERS`).
- `src/tools.ts` — pods (incl. start/stop via `POST /pods/{id}/action`), templates, network volumes, registries, and catalog (GraphQL→REST under v2) routed through `resolveBackend`; new `list-cpu-types` / `get-gpu-type`; descriptions + annotations (readOnly/destructive/idempotent) on all tools.
- `scripts/smoke-crud.ts` — live create→get→delete smoke (free resources only, fail-closed teardown).

2. **#45 — Route every resource through the adapter** (v1 + v2) ← you are here

|---|---|---|

Do not merge until v2 is live. Supersedes drafts #37, #38, #39, #41, #43.


src/_shared/mappers.ts
```diff
@@ -0,0 +1,175 @@
...
+  params: V1TemplateCreate
+): Record<string, unknown> {
+  return {
+    ...containerConfigToV2(params),
+    ...compact({
+      name: params.name,
+      serverless: params.isServerless,
+    }),
+    // Required by v2 — always emit, defaulting to NVIDIA.
+    category: params.category ?? 'NVIDIA',
```
LarveyOfficial, line 242: 🟠 High: this unconditionally emits category: 'NVIDIA' because v2 marks it required and the v1 tool schema exposes no category field — so the `category?` interface field is never populated and there's no override path. Every CPU/AMD template created via the MCP tool on v2 is silently mis-categorized as NVIDIA. Suggest exposing `category` on the create-template tool schema now (or inferring it from isServerless/image) rather than hardcoding.

src/_shared/mappers.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// gpuTypeIds[] + gpuCount → gpu: { id, count }. v2 GpuConfig requires `id`, so
+// return undefined for a count-only input — an idless gpu would be rejected
+// (422). count is optional (defaults to 1 server-side).
+function gpuConfigToV2(
+  gpuTypeIds?: string[],
+  gpuCount?: number
+): Record<string, unknown> | undefined {
+  const id = gpuTypeIds?.[0];
```
LarveyOfficial, line 81: 🟠 High: only the first GPU type is sent to v2; the rest are dropped silently. The v1 tool schema and fixtures accept arrays (e.g. ["NVIDIA A100","NVIDIA H100"]) — a caller passing a fallback list expecting "any of these" loses that semantic with no signal in the response. Either reject when length > 1 on v2, or document the truncation in the tool description and echo a warning in the reply.

src/_shared/mappers.ts
```diff
@@ -0,0 +1,175 @@
...
+  return compact({ id, count: gpuCount });
+}
+
+export function mapPodCreateToV2(params: V1PodParams): Record<string, unknown> {
+  return compact({
+    ...containerConfigToV2(params),
+    name: params.name,
+    cloud: params.cloudType,
+    // array → scalar: v2 takes a single GPU type id / data center
+    dataCenter: params.dataCenterIds?.[0],
```
LarveyOfficial: 🟠 High: same truncation as gpuTypeIds — only the first data center is sent on v2 and additional entries are dropped silently. Reject on length > 1 or document + warn.

src/tools.ts
```diff
@@ -236,7 +305,39 @@ export function registerTools(
           'Include GPUs that are currently out of stock. Default is false'
         ),
     },
+    { title: 'List GPU types', ...READ_ONLY },
     async (params) => {
+      const backend = backendFor('gpus');
+      if (backend.version === 'v2') {
+        // v2 REST: GET /v2/catalog/gpus → { gpus: [...] }. Filters re-applied
```
LarveyOfficial: 🟡 Medium: on v2 this does one catalog GET, unwraps, then filters and caps in memory — it never forwards/follows an upstream cursor. If the v2 catalog endpoint paginates server-side, this silently returns only the first server page (and the local capListResult cursor is unrelated to the upstream one). Please confirm the v2 catalog endpoints are non-paginated; if they can paginate, follow the upstream cursor before filtering.

src/tools.ts
```diff
@@ -236,7 +305,39 @@ export function registerTools(
...
+        // against v2 field names (memory/secure/community/name). v2 has no
+        // realtime stockStatus, so `includeUnavailable` is a documented no-op.
+        const raw = await callRestUrl(`${backend.base}${backend.list}`);
+        let gpus = backend.unwrap(raw) as Array<Record<string, unknown>>;
+        if (params.minMemoryGb !== undefined)
+          gpus = gpus.filter(
+            (g) => Number(g.memory ?? 0) >= params.minMemoryGb!
```
LarveyOfficial: 🟡 Medium: this assumes v2 returns `memory` in GB and `secure`/`community` as booleans. If v2 uses MiB, a different key (memoryInGb/vram), or an enum like availability:"secure", the filter silently passes everything or nothing. There's no fixture pinning the real v2 GPU shape — please pin the field names/units against the spec and handle the alternates defensively.

src/_shared/mappers.ts
```diff
@@ -0,0 +1,175 @@
...
+    gpu: gpuConfigToV2(params.gpuTypeIds, params.gpuCount),
+  });
+}
+
+export function mapPodUpdateToV2(params: V1PodParams): Record<string, unknown> {
+  // Update has no gpu/cloud/dataCenter; just name + ContainerConfig fields.
+  return {
+    ...containerConfigToV2(params),
+    ...compact({ name: params.name }),
```
LarveyOfficial, line 108: 🟡 Medium: an update with no mappable fields produces {} and is PATCHed as an empty body — please confirm v2 PATCH tolerates that (and add a test). Minor: containerConfigToV2 already compacts and `name` is compacted separately, which is inconsistent with mapPodCreateToV2's outer compact(...).

src/tools.ts
```diff
@@ -205,12 +230,56 @@ export function registerTools(
...
+  ): Promise<unknown> =>
+    withApiErrorLog('Error calling Runpod API:', () =>
+      restClient(url, method, body)
+    );
+  const backendFor = (
+    resource: Parameters<typeof resolveBackend>[0]['resource']
+  ): Backend =>
+    resolveBackend({
+      resource,
+      env: process.env as Env,
```
LarveyOfficial: 🟡 Medium: this unchecked cast is the intended call-time env seam, but a malformed value (e.g. RUNPOD_REST_VERSION=V2 uppercase, or "vtwo") depends entirely on how resolveVersion treats unknowns. Please confirm resolveVersion normalizes/rejects unrecognized values rather than silently falling through to v1 (or throwing at handler time).

src/_shared/mappers.ts
```diff
@@ -0,0 +1,175 @@
...
+): Record<string, unknown> {
+  return {
+    ...containerConfigToV2(params),
+    ...compact({ name: params.name }),
+  };
+}
+
+// Registry create is unchanged between v1 and v2 (name/username/password) →
+// identity. Exposed for symmetry so the descriptor table is uniform.
+export function identityMapper(body: unknown): unknown {
```
LarveyOfficial: 🔵 Low: identityMapper is exported "for symmetry" but backend.ts uses its own local `identity` and never imports this — it's dead. Either use it in backend.ts (one identity) or remove it.

LarveyOfficial (commented):
Automated review pass — read the full diff. High-quality work with strong wire-lock goldens guarding v1 behavior. Two things I'd treat as blockers are left inline (hardcoded NVIDIA template category; v2 dropping all-but-first GPU type / data center). A handful of mediums and nits inline as well. Verified correct: the /v2/v2 double-prefix is test-guarded; persistentMount drops partial mounts to avoid 422; mapTemplateUpdateToV2 maps imageName→image; podAction builds /action (v2) vs /start|/stop (v1); test beforeEach clears RUNPOD_REST_VERSION* env. No injection/secrets/auth issues found.

src/tools.ts
```diff
@@ -372,14 +492,57 @@ export function registerTools(
...
+    'List available CPU flavor types for CPU pods/endpoints. v2-only — returns a 501 notice on the v1 API.',
+    { ...listPaginationParams },
+    { title: 'List CPU types', ...READ_ONLY },
+    async (params) => {
+      const backend = backendFor('cpus');
+      if (backend.version === 'v1') {
+        return jsonReply({
+          error:
+            'list-cpu-types is only available on the v2 REST API. Set RUNPOD_REST_VERSION=v2.',
+          status: 501,
```
LarveyOfficial: 🔵 Low: on v1 these return a *successful* tool reply whose JSON contains status: 501 — an LLM client may not distinguish that from a real result. Consider setting isError: true on the MCP reply so clients treat it as a failure. (#46 expands the 501 approach.)

justinwlin:
Review findings on this PR's mappers are fixed in **commit d344305** (landed on the stack tip #47, since the per-resource split moved these files there; it's in the merged result and was rebased into #49's preview):

- **`mapPodCreateToV2` data centers (the real bug):** it emitted a singular `dataCenter` key, but the v2 `CreatePodRequest` field is **`dataCenterIds` (array)** — the singular key doesn't exist on the pod schema, so placement was silently dropped. Now maps to `dataCenterIds` (full array passthrough).
- **Multi-GPU:** v2 `gpu.id` is singular (so `[0]` is forced), but the narrowing was silent — create-pod now returns a `_warning` listing the ignored gpuTypeIds.
- **Template `category` hardcoded NVIDIA:** exposed an optional `category` on the create-template schema so non-NVIDIA templates aren't mis-categorized; it flows through the existing `?? 'NVIDIA'` default.

Mapper unit tests + the `v2-create-pod.json` fixture were updated to assert the corrected `dataCenterIds` shape.

src/_shared/mappers.ts
justinwlin: Fixed in d344305 — `create-template` now exposes an optional `category` (CPU|NVIDIA|AMD) that flows through the mapper's existing `?? 'NVIDIA'` default, so non-NVIDIA templates are no longer silently mis-categorized.

src/_shared/mappers.ts
justinwlin: Fixed in d344305 — v2's `gpu.id` is genuinely singular (spec `GpuConfig`), so `[0]` is forced, but it's no longer silent: create-pod echoes a `_warning` naming the ignored gpuTypeIds. + handler test.

src/_shared/mappers.ts
justinwlin: Fixed in d344305 — and it was worse than truncation: the singular `dataCenter` key doesn't exist on the v2 `CreatePodRequest` (the field is `dataCenterIds`, an **array**, spec L943), so DC placement was dropped entirely on v2 pod create. Now maps `dataCenterIds` as a full-array passthrough (dropped when empty). Fixture + unit/handler tests updated.

src/tools.ts
justinwlin: Confirmed non-paginated — `ListGpuTypesResponse` is `{ gpus: [...] }` with no cursor field, and the live `/v2/catalog/gpus` returns a flat list. The in-memory filter/cap is correct here.

src/tools.ts
justinwlin: Confirmed against the live v2 API: `/v2/catalog/gpus` returns `memory` as a number in GB and `secure`/`community` as booleans, so the filter's assumptions hold. A pinned GpuType fixture would be good belt-and-suspenders — noted as a follow-up.

src/_shared/mappers.ts
justinwlin: Noted (low blast, read/PATCH path) — an empty PATCH body is a harmless no-op update; a guard + test is a reasonable follow-up. The `name`/compact inconsistency is cosmetic.

src/tools.ts
justinwlin: Confirmed — `resolveVersion` lowercases the value (so `V2`→`v2` works) and any unrecognized value (`vtwo`) falls through to `v1`. It never throws on a bad value, and a missing env defaults to `v1` before `.toLowerCase()`.

src/_shared/mappers.ts
justinwlin: Fixed in 2d8685d — removed the dead `identityMapper` export (backend.ts uses its own local `identity`; nothing referenced this).

src/tools.ts
justinwlin: Noted — the readable 501 notice is intentional; standardizing the v2-only notices on `isError:true` is a reasonable follow-up across the whole 501 surface (expanded in #46).

src/tools.ts
LarveyOfficial: Checked resolveVersion: case is normalized via .toLowerCase() (so V2 works), but an unrecognized value (e.g. a typo'd vtwo) still falls through to return 'v1' silently, with no warning. Safe default, but a misconfiguration would be invisible. Leaving open in case you want to warn/reject on unknown values.

src/_shared/mappers.ts
LarveyOfficial: @justinwlin 
Update: the mapper now accepts an override (category: params.category ?? 'NVIDIA') — good groundwork. But the create-template tool schema still exposes no category field, so params.category is never populated and templates created through the MCP tool are still forced to NVIDIA. Leaving open until the tool surfaces category (or infers it from image/serverless).

scripts/smoke-crud.ts
jebenexer, line 1: **Two issues in this file:**

**1. `RUNPOD_REST_V2_API_URL` never reaches the child process**

The explicit `env` object passed to `StdioClientTransport` **replaces** (not merges) the child process environment — `process.env` is not inherited. The usage comment instructs callers to set `RUNPOD_REST_V2_API_URL=https://v2-rest.runpod.dev/v2`, but the child never receives it, so the v2 smoke run silently validates against production `v2-rest.runpod.io`. `PATH` is also stripped, which may prevent the child from finding `node` depending on the OS.

The fix added to `scripts/smoke.ts` in #46 handles this correctly. Apply the same pattern here:
```ts
const childEnv: Record<string, string> = { RUNPOD_API_KEY: apiKey(), RUNPOD_REST_VERSION: version };
for (const [k, v] of Object.entries(process.env)) {
  if (v !== undefined && (k.startsWith('RUNPOD_REST') || k === 'RUNPOD_SERVERLESS_API_URL' || k === 'RUNPOD_PUBLIC_GRAPHQL_URL'))
    childEnv[k] = v;
}
```

**2. Unknown version arguments silently dropped → false green exit 0**

```ts
const versions = process.argv.slice(2).filter((v) => v === 'v1' || v === 'v2');
if (versions.length === 0) versions.push('v1');
```

A typo like `V2` or `v2 ` is silently discarded and the script exits 0 after running only v1. For a gating script this is a dangerous false green. Add:
```ts
const bad = process.argv.slice(2).filter(v => v !== 'v1' && v !== 'v2');
if (bad.length) { console.error(`Unknown version(s): ${bad.join(', ')}`); process.exit(1); }
```

jebenexer (changes requested):
Two blockers in `smoke-crud.ts`. Both prevent the v2 smoke from validating what it claims to validate — the child process drops the v2 base URL override and silently hits production, and unknown version args produce a false green exit.

justinwlin:
Thanks @jebenexer — both `smoke-crud.ts` blockers fixed (on the stack tip `v2/group-4-source-split`, commit `29361d0`, to avoid a 4-branch rebase; present in the merged result):

1. **Child env replaced → v2 smoke hit prod.** Now forwards `RUNPOD_REST*` / `RUNPOD_SERVERLESS_API_URL` / `RUNPOD_PUBLIC_GRAPHQL_URL` into the child (same shape as `smoke.ts`), but keeps `RUNPOD_REST_VERSION` authoritative to each run's `version` arg (so a parent value can't override the per-version loop). The dev base now actually reaches the child.
2. **Unknown version args → false green.** Now rejects anything other than `v1`/`v2` with a non-zero exit instead of silently dropping it.

scripts/smoke-crud.ts
justinwlin: Both fixed in `29361d0` (on the tip `v2/group-4-source-split`):

1. **Child env replaced → v2 smoke hit prod.** Now forwards `RUNPOD_REST*` / `RUNPOD_SERVERLESS_API_URL` / `RUNPOD_PUBLIC_GRAPHQL_URL` into the child (same shape as `smoke.ts`), but keeps `RUNPOD_REST_VERSION` authoritative to each run's `version` arg so a parent value can't override the per-version loop. The dev base now actually reaches the child.
2. **Unknown version args → false green.** Now rejects anything other than `v1`/`v2` with a non-zero exit instead of silently dropping it.

jebenexer (approved):
Both `smoke-crud.ts` blockers verified fixed in 29361d0: child env now forwards `RUNPOD_REST*` vars (RUNPOD_REST_VERSION kept authoritative per-run so the parent value can't override the loop), and unknown version args now exit non-zero instead of silently falling back. The gating script is reliable.

PR #40, opened 2026-06-23
[Phase C · PR7] restart-pod (C1) + create-pod 501 clean error (C2)

justinwlin:
**Stacked draft — do not merge.** Base = `v2/phase-b-crud-routing` (PR #39). Implements **C1 (restart)** + **C2 (501 handling)**.

## Changes
- **C1 `restart-pod`** — new tool via the existing `podAction` helper. v2 → `POST /v2/pods/{id}/action {action:"restart"}` (PodAction enum includes `restart`); v1 → `POST /pods/{id}/restart` (v1 has no restart endpoint — documented, best-effort).
- **C2 create-pod 501** — catches `HttpError` with `status === 501` (v2 stubs CPU-pod creation, AE-2991) and returns a **clean non-error message** ("CPU pods are not yet supported on the v2 REST API…") instead of a raw throw. Other errors still propagate; `stream-job`'s consecutive-error counter is unaffected (the client still throws).

## Tests
- restart-pod: registration + v1 (`/pods/{id}/restart`) + v2 (`/action {action:"restart"}`).
- create-pod 501 → asserts non-error content + clean message (not a throw).
- **153 pass** · `tsc`/`eslint`/`tsup` clean.

## Scope
`list-cpu-types` / `get-gpu-type` (the other two C1 tools) ship with the **B5 catalog GraphQL→REST** PR, since they require catalog v2 REST routing. C3 (probe wiring + flip default) and B6 (live CRUD smoke) follow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #46, opened 2026-06-24
[v2 3/4] New capabilities (restart, 501, auto-probe)

justinwlin:
**Part 3 of 4 — the v2 REST migration.**

## This PR: New capabilities (v2-only behavior)
- `restart-pod` tool (v2 `PodAction`; clean 501 notice under v1).
- create-pod **501** handling → clean "CPU pods not yet supported on v2" message instead of a raw error.
- `auto` mode wiring in `stdio.ts`: probe v2 once at startup, fall back to v1; typed against the exported `FetchLike` (no `Parameters<…>` cast).
- README brought up to date for v2 (version env vars, new tools, dev commands).

3. **#46 — New capabilities** (restart-pod, 501, auto-probe) ← you are here

|---|---|---|

Do not merge until v2 is live. Supersedes drafts #40, #42.


justinwlin:
## v2 REST migration — manual validation pass

Ran the full playbook below against the **v2 dev** API (`v2-rest.runpod.dev/v2`) on branch `v2/group-3-capabilities`, driving each tool over the local HTTP MCP server one row at a time. **All 26 rows PASS**; teardown clean (no `mcptest-*` leaks).

> Evidence is **API-level** (returned ids, `status` fields, `204` deletes, end-of-run leak sweep) — not console-UI screenshots.

## Results

| Step | Tool | Result | Notes |
|------|------|--------|-------|
| A1 | list-gpu-types | PASS | `{items,pagination}` envelope; 5 types @ `minMemoryGb:24`; cheapest in-stock community = RTX 4090 ($0.64/hr) |
| A2 | list-data-centers | PASS | envelope, `truncated:true` + `nextCursor` + note; **cursor page-2 verified** (offset 20, 3 items); NV-capable DCs: CPU-2, MFS-1 |
| A3 | list-cpu-types | PASS | envelope, 6 CPU flavors (CPU3/CPU5) — v2-only |
| A4 | get-gpu-type | PASS | A40 matches A1 exactly |
| B1 | create-template | PASS | id `ru5n1lipag`, image `python:3.11-slim` |
| B2 | get-template | PASS | round-trips id + image |
| B3 | list-templates | PASS | mcptest-tpl present |
| B4 | update-template | PASS | name → mcptest-tpl-2 |
| B5 | delete-template | PASS | `204` |
| B6 | list-templates (gone) | PASS | mcptest-* absent |
| C1 | create-registry-auth | PASS | id `cmqsgvkd1000gktxpenl2wvpk` |
| C2 | get-registry-auth | PASS | metadata only — **no secret returned** |
| C3 | delete-registry-auth | PASS | `204` |
| D1 | create-network-volume | PASS | id `h9h6518urm`, CPU-2 (DC chosen dynamically from A2) |
| D2 | get-network-volume | PASS | size 10, CPU-2 |
| D3 | update-network-volume | PASS | grew 10 → 20 |
| D4 | delete-network-volume | PASS | `204` |
| E1 | create-pod | PASS | id `u79hehpx250uj3`, RUNNING, MFS-1, $0.74/hr |
| E2 | get-pod | PASS | RUNNING |
| E3 | stop-pod | PASS | EXITED (actions → start/terminate) |
| E4 | start-pod | PASS | RUNNING again |
| E5 | restart-pod | PASS | RUNNING, `startedAt` advanced (v2 PodAction) |
| E6 | delete-pod | PASS | `204`, teardown ok |
| F | v1↔v2 parity | PASS (reads) | A1/A2 reads work on v1; both versions now wrap in the pagination envelope. Inner fields differ as expected: v1 raw (`displayName`, `memoryGb`, `stockStatus`) vs v2 normalized (`manufacturer`, `maxCount`, `pool`, `price`). **B1–B6 writes out of scope** — dev key 401s on v1-prod (per `test.md` caveat). |
| G1 | list-cpu-types on v1 | PASS | clean 501 — "only available on the v2 REST API. Set RUNPOD_REST_VERSION=v2." |
| G2a | get-gpu-type on v1 | PASS | clean 501 + hint ("use list-gpu-types on v1") |
| G2b | restart-pod on v1 | PASS | clean 501 + hint ("stop then start the pod on v1"); tool now **registered** on v1 (v1 = 39 tools) |
| G3 | CPU pod on v2 | PASS | GPU-less create transparently routed to v1 (`tools.ts:717`, tags `_servedBy:'v1'`). Dev key → `401` from v1-prod = expected; confirms **routing** (was a raw v2 `400 "exactly one of gpu or cpu"` before the fix). |

**Leak check (end of run):** `list-templates`, `list-network-volumes`, `list-container-registry-auths`, `list-pods` → **NONE `mcptest-*` remain** across all four.

---

## Sign-off

- **#44** Adapter foundation — v1 unchanged, server boots both versions, handshake 200 OK: **✓**
- **#45** Routing — A–E + F reads + G all pass on v2: **✓**
- **#46** Capabilities — E5 restart, G1–G3 notices, G3 v1-routing: **✓**

---

## Notes / observations (not blockers)

1. **Catalog lists are now paginated.** `list-gpu-types`, `list-data-centers`, `list-cpu-types` return `{ items, pagination }` (default 20). `pagination` carries `total / returned / offset / truncated / nextCursor` (+ a human `note` when truncated). Cursor paging confirmed working (A2 page 2).
2. **Both versions wrap in the same pagination envelope**, but inner item shapes differ between v1 (raw GraphQL-ish fields) and v2 (normalized REST fields). Expected — the adapter normalizes v2, not v1.
3. **v1↔v2 switch is a server-process change**, done by relaunching the launcher with/without `RUNPOD_REST_VERSION`. v1 tool count = 39 (now includes `restart-pod` as a 501-gated entry); v2 = 39.
4. **All prior-round findings resolved:** G3 CPU-pod guard (now routes to v1), restart-pod gating inconsistency (now registered + clean 501 on v1), and the playbook data bugs (fake `runpod/pytorch:1.0.2` image, hardcoded `US-NJ` DC, missing F dev-key caveat).

## Reproducing

```bash
# v2 dev server
RUNPOD_REST_VERSION=v2 RUNPOD_REST_V2_API_URL=https://v2-rest.runpod.dev/v2 PORT=3399 \
  npx tsx scripts/serve-http.local.ts

# drive one row (uncommitted helper)
RUNPOD_API_KEY=rpa_YOUR_DEV_KEY npx tsx scripts/call.local.ts list-gpu-types '{"minMemoryGb":24}'

# v1 (for F / G1 / G2): relaunch without RUNPOD_REST_VERSION
PORT=3399 npx tsx scripts/serve-http.local.ts
```

---

<details>
<summary><b>Full test plan (playbook driven for this run)</b></summary>

# Runpod MCP — v2 manual validation playbook

A guided, human-in-the-loop test pass for the v2 REST migration (PRs **#44 → #45 → #46**).
You launch the server locally pointed at the **dev** API, an agent drives the tools
step by step, and **you** confirm in the Runpod console that things really spin up,
update, and get torn down. Results go in the table at the bottom and become the
validation evidence for the PRs.

> ⚠️ This hits a **real dev account**. Some steps create billable resources (pods,
> network volumes). The agent is instructed to **pause and ask before anything that
> costs money**, and every created resource is named `mcptest-*` and deleted at the end.

---

## 0. Prereqs

- A Runpod **dev-account** API key → `rpa_YOUR_DEV_KEY` below.
- The dev REST URL: `https://v2-rest.runpod.dev/v2`.
- A clone of this repo **on the v2 branch** with deps installed: `git checkout v2/group-3-capabilities && pnpm install` (a fresh clone defaults to `main`, which has none of this).
- An MCP-capable agent (Claude Code, Claude Desktop, Cursor, …).

**Key / environment matrix** (which key works where):

| Key | Authenticates against | Use for |
|-----|----------------------|---------|
| dev-account key | dev v2 (`v2-rest.runpod.dev/v2`) | **this walkthrough** — set as the Bearer |
| prod-account key | v1 prod (`rest.runpod.io/v1`) | v1 CRUD parity (section F) — mutates a **real prod account**, be careful |

Use the **dev** key here. v2 **prod** (`v2-rest.runpod.io/v2`) is reachable but its auth isn't cut over yet, so a prod key currently fails there.

---

## a) Launch the server locally

The server runs in **HTTP** mode so an agent can connect to a URL. The caller's
API key is sent as the Bearer token; v2 + the dev host are set on the server process.

### 1. Make sure the local HTTP launcher exists

It's an uncommitted dev helper. If `scripts/serve-http.local.ts` is missing, create it:

```ts
// scripts/serve-http.local.ts
import { createServer } from 'node:http';
import { handleMcpRequest } from '../src/http.js';

const PORT = Number(process.env.PORT ?? 3399);
const server = createServer((req, res) => {
  handleMcpRequest(req, res, { serverVersion: 'local-dev' }).catch((err) => {
    console.error('mcp handler error:', err);
    if (!res.headersSent) {
      res.writeHead(500, { 'Content-Type': 'application/json' });
      res.end(JSON.stringify({ error: 'internal error' }));
    }
  });
});
server.listen(PORT, () =>
  console.error(
    `MCP HTTP server on http://localhost:${PORT}/ ` +
      `(RUNPOD_REST_VERSION=${process.env.RUNPOD_REST_VERSION}, base=${process.env.RUNPOD_REST_V2_API_URL})`
  )
);
```

### 2. Start it (v2, dev host)

```bash
# from the repo root
RUNPOD_REST_VERSION=v2 \
RUNPOD_REST_V2_API_URL=https://v2-rest.runpod.dev/v2 \
PORT=3399 \
npx tsx scripts/serve-http.local.ts
```

You should see `MCP HTTP server on http://localhost:3399/`.

Sanity checks (in another terminal):

```bash
# no auth → 401 (expected)
curl -s -o /dev/null -w '%{http_code}\n' http://localhost:3399/ \
  -H 'Accept: application/json, text/event-stream' -d '{}'

# handshake → 200 + serverInfo
curl -s http://localhost:3399/ \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H 'Authorization: Bearer rpa_YOUR_DEV_KEY' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"verify","version":"0"}}}'
```

### 3. Connect your agent

**Claude Code:**

```bash
claude mcp add --transport http runpod-v2 http://localhost:3399/ \
  --header "Authorization: Bearer rpa_YOUR_DEV_KEY"
```

**Generic (Cursor/Desktop-style):**

```json
{
  "mcpServers": {
    "runpod-v2": {
      "url": "http://localhost:3399/",
      "headers": { "Authorization": "Bearer rpa_YOUR_DEV_KEY" }
    }
  }
}
```

> To compare against **v1**, stop the server and relaunch without `RUNPOD_REST_VERSION`
> (or `=v1`). Same key, same agent — only the version changes.

---

## b) Drive the test with an agent (copy-paste prompt)

Paste this to the agent connected to `runpod-v2`. It walks the suite **one step at a
time**, pausing for you to confirm in the console, and asks before anything billable.

> You are connected to the `runpod-v2` MCP server (Runpod v2 REST, dev account).
> Walk me through the validation suite in `test.md`, **one numbered row at a time**
> (A1, A2, … B1, B2, … — each *row* is a step, not each section). Before you start,
> read the **"Agent rules"** section just below this prompt and follow it throughout.
> For each row:
> 1. Tell me which tool you're about to call and with exactly what arguments.
> 2. If the row is a **create / update / delete** (or costs money), **stop and wait for
>    my explicit "go"** — one fresh "go" per mutating row; never chain two mutating
>    rows on one approval. Read-only rows you may run directly.
> 3. After the call, show me the raw result. If the row has a console check, tell me
>    exactly **what to look for and where** in the console (https://console.runpod.io).
>    If the check is "—", say "no console check — confirming via tool output" and show
>    the proving field (e.g. the returned id).
> 4. Wait for my "verified" before continuing.
> 5. Record the row as PASS / FAIL / NEEDS-ATTENTION / BLOCKED with a one-line note.
> Name every created resource `mcptest-<row>`. When a create returns an id, record it
> and reuse it for that section's get/update/delete rows; keep a running list of live
> ids. At the end, fill the results table (header + rows + sign-off), print it, and
> list anything left behind. Do not invent results — only report actual tool output.

### Agent rules (failure, teardown, cost, IDs) — read before starting

- **Teardown is mandatory.** If you received an id for any billable resource (pod,
  network volume), you MUST run its delete row before ending — even if intermediate
  steps failed or I abort. Confirm the deletion. A leaked pod/volume keeps billing.
- **Cost approval.** Before creating a pod or volume, show me the concrete config and
  its price in your "go" prompt (for a pod: the chosen GPU type + its per-hour price
  from A1) so I approve the real cost. Default to the cheapest in-stock option.
- **Create fails →** skip that section's dependent get/update/delete rows, mark them
  BLOCKED, and move to the next section (no id means nothing to act on).
- **Delete returns 404 / not-found →** that's a **PASS**: deletes are idempotent
  (re-deleting a gone resource is a no-op).
- **Delete tool errors otherwise →** tell me to delete `<resource> <id>` manually in
  the console at the listed location, and mark it NEEDS-ATTENTION / leak.
- **Pod not RUNNING →** poll `get-pod` ~every 15s for up to ~3 minutes. If it never
  reaches RUNNING (e.g. no capacity), mark NEEDS-ATTENTION and **still run E6 delete**
  so nothing leaks.
- **Negative cases (G1–G3):** a clean "v2 only" / 501 notice **IS the PASS** condition;
  a raw crash or stack trace is the FAIL.
- **Switching v1 ↔ v2 (sections F, G1, G2)** is a **server-process** change, not a tool
  arg. Pause and tell me to stop the server and relaunch it without `RUNPOD_REST_VERSION`
  (= v1), then reconnect — you cannot do this yourself.

---

## c) The test suite (steps + manual checks)

Tip: a name prefix of `mcptest-` makes leaks obvious. Free resources first, billable last.

### A — Catalog (read-only, safe, no teardown)

> Catalog lists are now **paginated** like the other list tools — the result is a
> `{ items, pagination }` envelope (default 20 items). Read `items`; if
> `pagination.truncated` is true, pass `cursor: pagination.nextCursor` for the next page.

| # | Tool | Args | Expect | Manual check |
|---|------|------|--------|--------------|
| A1 | `list-gpu-types` | `minMemoryGb: 24` | list of GPU types (v2 fields) | results look sane |
| A2 | `list-data-centers` | — | list of DCs with regions | pick a `dataCenterId` whose `networkVolumeTypes` is non-empty (e.g. offers CPU-2 / MFS-1) **and** is in stock — use it for D1; don't hardcode a DC (they go out of capacity) |
| A3 | `list-cpu-types` | — | CPU flavors (v2-only) | non-empty on v2 |
| A4 | `get-gpu-type` | an id from A1 | one GPU type | matches A1 |

### B — Templates (free; full CRUD)
| # | Tool | Args | Expect | Manual check |
|---|------|------|--------|--------------|
| B1 | `create-template` | `name: mcptest-tpl`, `imageName: python:3.11-slim`, `containerDiskInGb: 10`, `isServerless: false` | returns an `id` | — |
| B2 | `get-template` | `templateId` from B1 | round-trips id + image | — |
| B3 | `list-templates` | — | B1 present | template shows in console → Templates |
| B4 | `update-template` | new `name: mcptest-tpl-2` | reflects new name | — |
| B5 | `delete-template` | `templateId` | success | gone from console |
| B6 | `list-templates` | — | B1 **absent** | confirm no `mcptest-*` remain |

### C — Container registry auth (free; CRUD)
| # | Tool | Args | Expect | Manual check |
|---|------|------|--------|--------------|
| C1 | `create-container-registry-auth` | `name: mcptest-reg`, `username: x`, `password: y` | returns `id` | appears in console → Settings → Container Registry |
| C2 | `get-container-registry-auth` | id | metadata (no secret) | — |
| C3 | `delete-container-registry-auth` | id | success | gone from console |

### D — Network volumes (⚠️ billable storage; CRUD) — ask before D1
| # | Tool | Args | Expect | Manual check / intervention |
|---|------|------|--------|------------------------------|
| D1 | `create-network-volume` | `name: mcptest-vol`, `size: 10`, `dataCenterId:` from A2 | returns `id` | volume appears in console → Storage; **billing starts** |
| D2 | `get-network-volume` | id | size + DC | — |
| D3 | `update-network-volume` | `size: 20` | grown | size updated in console |
| D4 | `delete-network-volume` | id | success | **gone** from console; billing stops |

### E — Pods (⚠️ billable GPU/CPU; full lifecycle) — ask before E1
| # | Tool | Args | Expect | Manual check / intervention |
|---|------|------|--------|------------------------------|
| E1 | `create-pod` | `name: mcptest-pod`, `imageName: python:3.11-slim`, `gpuTypeIds: [<cheapest in-stock community GPU id from A1>]`, `gpuCount: 1`, `containerDiskInGb: 10` | returns `id` | agent shows chosen GPU + price for approval; pod appears in console → Pods; **poll get-pod until RUNNING** |
| E2 | `get-pod` | id | status transitions to running | confirm boot + logs in console |
| E3 | `stop-pod` | id | stopped | console shows stopped; **GPU billing stops**, disk persists |
| E4 | `start-pod` | id | running again | console shows running |
| E5 | `restart-pod` | id | restarts (v2 PodAction) | console shows restart |
| E6 | `delete-pod` | id | success | **gone** from console; billing fully stops |

### F — v1 ↔ v2 parity (relaunch server on v1, repeat a subset)
Re-run **A1, A2, B1–B6** under v1 (`RUNPOD_REST_VERSION` unset). Expect the same
behavior and shapes. Note any difference.

> **Key caveat:** v1 has only a **prod** base (`rest.runpod.io/v1`). The dev key does
> catalog reads (A1/A2 are public) but **401s on v1 writes** — so B1–B6 under v1 need
> either a **prod** key (mutates a real prod account — be careful) or `RUNPOD_REST_API_URL`
> pointed at a v1 dev endpoint. With only the dev key, limit F to the read-only A1/A2.

### G — Negative / edge cases
| # | Check | How | Expect |
|---|-------|-----|--------|
| G1 | v2-only tool under v1 | run `list-cpu-types` with server on v1 | clean "v2 only" notice (status 501), not a crash |
| G2 | v2-only tool under v1 | `get-gpu-type` / `restart-pod` on v1 | same clean notice |
| G3 | CPU pod on v2 → routes to v1 | `create-pod` with `imageName`, `containerDiskInGb`, **no** `gpuTypeIds`, on v2 | reply is flagged `_servedBy:"v1"` (transparently routed to v1 — not a raw v2 400). ⚠️ with the **dev** key the v1-prod write then **401s** — that's expected; G3 verifies the *routing*, not a successful pod. If it does succeed it created a real CPU pod on v1 → **delete it** (`delete-pod`). |

---

## Results (fill in — this is the PR validation record)

Server: version `____`  base `____`  date `____`  tester `____`

| Step | Tool | Result (PASS/FAIL/ATTN) | Notes / console observation |
|------|------|--------------------------|------------------------------|
| A1 | list-gpu-types | | |
| A2 | list-data-centers | | |
| A3 | list-cpu-types | | |
| A4 | get-gpu-type | | |
| B1 | create-template | | |
| B2 | get-template | | |
| B3 | list-templates | | |
| B4 | update-template | | |
| B5 | delete-template | | |
| B6 | list-templates (gone) | | |
| C1 | create-registry-auth | | |
| C2 | get-registry-auth | | |
| C3 | delete-registry-auth | | |
| D1 | create-network-volume | | |
| D2 | get-network-volume | | |
| D3 | update-network-volume | | |
| D4 | delete-network-volume | | |
| E1 | create-pod | | |
| E2 | get-pod (RUNNING) | | |
| E3 | stop-pod | | |
| E4 | start-pod | | |
| E5 | restart-pod | | |
| E6 | delete-pod | | |
| F | v1↔v2 parity | | |
| G1–G3 | negative cases | | |

**Leak check (run at the end):** `list-templates`, `list-network-volumes`,
`list-container-registry-auths`, `list-pods` → confirm **no `mcptest-*` remain**.
(The automated equivalent for free resources is `pnpm smoke:crud v1 v2`.)

**Sign-off:**
- #44 Adapter foundation — v1 unchanged, server boots, handshake OK: `____`
- #45 Routing — A–D + F + G all pass on v2: `____`
- #46 Capabilities — E5 restart, G1–G3 notices: `____`

---

## Saving results to the PRs

Once filled, attach this as evidence:

```bash
# paste the filled results table as a comment on each PR
gh pr comment 45 --body-file test.md      # or a trimmed results-only version
```

Or commit `test.md` (with results) onto the relevant branch if you want it tracked.

</details>

src/tools.ts
```diff
@@ -708,13 +719,75 @@ export function registerTools(
...
+            error:
+              'A GPU pod needs gpuTypeIds (see list-gpu-types); gpuCount or computeType alone is under-specified.',
+            status: 400,
+          });
+        }
+        // v2 has no CPU pods yet (AE-2991): serve a CPU pod from v1 (which supports
+        // them) instead of failing, and tell the caller which API answered.
+        if (wantsCpu) {
+          const result = (await callRestUrl(
+            `${restV1Base(process.env as Env)}/pods`,
```
LarveyOfficial: 🟠 High: the CPU v2→v1 fallback returns *before* the try/catch that converts HttpError/501 into a clean reply — so the new error handling only protects the GPU/v2 path. Any HttpError from this CPU create (400/401/500) surfaces as a raw thrown error/stack, which is exactly what this PR set out to avoid. It also passes the raw `params` (including computeType) straight through as the v1 body rather than running backend.mapCreate; if v1 rejects unknown fields or expects a mapped shape this will fail. Suggest routing the CPU fallback through the same error handling and deciding deliberately whether v1 gets mapped vs raw params.

src/_shared/backend.ts
```diff
@@ -132,6 +132,15 @@ export function resolveVersion(opts: {
   return 'v1';
 }
 
+// True if any version setting (global or per-resource) is `auto` — i.e. the
+// stdio entrypoint should run the v2 probe once at startup. Pure.
+export function wantsAutoProbe(env: Env): boolean {
+  return Object.entries(env).some(
+    ([k, v]) =>
+      k.startsWith('RUNPOD_REST_VERSION') && (v ?? '').toLowerCase() === 'auto'
```
LarveyOfficial, line 156: 🟡 Medium: this over-matches any RUNPOD_REST_VERSION* var, so a var like RUNPOD_REST_VERSION_ENDPOINTS=auto fires the startup probe (a network round-trip + potential ~4s hang) even though endpoints/jobs always route to v1 and can never be v2. Match RUNPOD_REST_VERSION exactly or only known-resource suffixes so the probe trigger and the routing stay consistent.

src/_shared/backend.ts
```diff
@@ -163,12 +181,13 @@ function classifyProbeStatus(status: number): boolean | undefined {
...
 
 // One probe attempt. Returns a definitive verdict, or `undefined` when the
-// attempt was transient (network throw / 5xx) and should be retried.
+// attempt was transient (network throw / timeout / 5xx) and should be retried.
 async function runV2Probe(deps: ProbeDeps): Promise<boolean | undefined> {
   try {
     const res = await deps.fetch(`${deps.baseUrl}/catalog/gpus`, {
       method: 'GET',
       headers: { Authorization: `Bearer ${deps.apiKey}` },
+      signal: AbortSignal.timeout(deps.timeoutMs ?? PROBE_TIMEOUT_MS),
```
LarveyOfficial, line 221: 🟡 Medium: the probe relies on AbortSignal.timeout(), but node-fetch@2 doesn't reliably honor WHATWG AbortSignal (and v3 has had abort-interop quirks). If the installed node-fetch ignores the signal, the "bound the startup probe so a slow host can't hang stdio" guarantee silently doesn't hold in production — and the unit test wouldn't catch it (the fake fetch only checks `signal instanceof AbortSignal`, never real cancellation). Please verify against the actual node-fetch version, or use Node 18+ global fetch / an explicit timeout wrapper, and add a test asserting the request is actually aborted.

src/_shared/backend.ts
```diff
@@ -163,12 +181,13 @@ function classifyProbeStatus(status: number): boolean | undefined {
 }
 
 // One probe attempt. Returns a definitive verdict, or `undefined` when the
-// attempt was transient (network throw / 5xx) and should be retried.
+// attempt was transient (network throw / timeout / 5xx) and should be retried.
```
LarveyOfficial, line 215: 🔵 Low: a timed-out/aborted probe degrades silently to v1, so a momentarily slow-but-reachable v2 host pins the whole session to v1 with no signal to the user. Consider logging when the probe times out so the fallback isn't invisible.

src/tools.ts
```diff
@@ -708,13 +719,75 @@ export function registerTools(
...
+            'POST',
+            params as Record<string, unknown>
+          )) as Record<string, unknown>;
+          return jsonReply({
+            ...result,
+            _servedBy: 'v1',
```
LarveyOfficial: 🔵 Low: spreading injected _-prefixed keys into the upstream pod object is fine for an LLM, but it collides if the real API response ever has _servedBy/_note, and leaks non-API fields to any programmatic consumer. Consider nesting under a single _meta key.

LarveyOfficial (commented):
Automated review pass — read the full diff. The create-pod validation matrix and the 501/restart handling are well-covered by the new golden tests. One item I'd treat as a blocker is inline (the CPU v2→v1 fallback skips the new error handling and sends raw params), plus a couple of mediums around the startup probe. Verified correct: the create-pod matrix (contradiction→400, under-specified GPU→400, absence→400, CPU→v1) with "fires no request" assertions; non-501 errors (400/500) correctly propagate; restart-pod on v1 returns a clean notice with zero outbound requests; 5xx kept transient/uncached in the probe.

justinwlin:
Re: the CPU-pod v2→v1 fallback error-handling finding — fixed in **commit d344305** (on the stack tip #47).

On investigation the impact was small (the existing `try/catch` only converts a 501, which v1 won't emit for a CPU pod; v1's `mapCreate` is the identity function so "raw params" == mapped; and the path was already proven working live). Tidied for consistency anyway: the CPU fallback now runs inside a try/catch and returns a clean `{error, status, _servedBy:'v1', _note}` reply on an `HttpError` instead of a raw throw. Added a handler test for that error path.

src/tools.ts
justinwlin: Fixed in d344305 — the CPU v2→v1 fallback now runs inside a try/catch and returns a clean `{ error, status, _servedBy:'v1', _note }` reply on `HttpError` instead of a raw throw. + handler test for the error path. (Re: raw params — for a v1 backend `mapCreate` is the identity function and `computeType` is a valid v1 field, confirmed by a live create→delete, so raw == mapped there; left as-is deliberately.)

src/_shared/backend.ts
justinwlin: Valid — noted as a follow-up. `wantsAutoProbe` over-matches `RUNPOD_REST_VERSION_*`, so a per-resource `=auto` on a v1-only resource (endpoints/jobs) would fire a needless probe. Low impact (rare misconfig); will tighten to known-resource suffixes.

src/_shared/backend.ts
justinwlin: Confirmed — we're on `node-fetch@3` (`^3.3.2`), which honors WHATWG `AbortSignal`/`AbortSignal.timeout`, so the probe timeout IS effective (the abort-interop issue was node-fetch@2). Agreed the only gap is a real-cancellation test — tracked as a follow-up.

src/_shared/backend.ts
justinwlin: Noted — good call; will add a log line when the probe times out so the v1 fallback isn't invisible.

src/tools.ts
justinwlin: Noted — fine for the LLM today; nesting the injected fields under a single `_meta` key (to avoid collisions / leaking non-API fields to programmatic consumers) is a reasonable follow-up.

LarveyOfficial:
@justinwlin 
I see the new test documents this as an intentional over-match ("matches any suffix … harmless"). Agreed it's harmless for correctness — the residual concern is cost: RUNPOD_REST_VERSION_ENDPOINTS=auto still triggers the startup probe (a network round-trip + up to the ~4s timeout) for a resource that always routes to v1. Not blocking; flagging so trigger/routing stay aligned if probe latency ever matters.

jebenexer (approved):
Reviewed the full diff. The existing review caught all the meaningful issues; two items worth noting before this lands:

**1. Probe timeout silent fallback — log line still missing (open thread PRRT_kwDOOP1NwM6MnWpt)**

Justin agreed to add a log when the probe times out and falls back to v1. It's not in 29361d0 and not visible in this branch HEAD. A timed-out probe silently pins the session to v1 with no signal — the log is a low-cost, high-value addition for any operator debugging unexpected v1 behavior. Please add before the stack merges.

**2. CPU fallback try/catch is on the stack tip, not this branch**

The High finding (LarveyOfficial) about the CPU v1 fallback missing a try/catch is marked resolved via d344305, but that fix is on `v2/group-4-source-split` (#47), not rebased into this branch. If #46 were merged alone, a CPU pod create that gets a 400/500 from v1 would surface as a raw thrown error. Given the ordered-merge strategy and the "do not merge until v2 is live" gate, #47 will land before any user sees #46 in isolation — so this is acceptable. Just confirming the fix is tracked and present in the stack.

Everything else looks solid: the intent-detection logic is correct and exhaustive, `restart-pod` fires no request under v1, the `AbortSignal.timeout` probe is verified against node-fetch@3, and the stdio async startup is properly awaited with a clean error path.

src/_shared/backend.ts
justinwlin: Done in `7b6c546` (on the stack tip `v2/group-4-source-split`). Added an optional injected `warn` callback to `ProbeDeps`; `runV2Probe` calls it on a **transient** fallback (timeout / network error / 5xx / 401-403) but **not** on a definitive 404 (v2 genuinely not deployed — the expected answer, not a blip). `stdio.ts` passes `console.error` (the MCP stderr log channel), so an `auto` session that silently lands on v1 now logs e.g.:

> `[runpod-mcp] v2 auto-probe: https://v2-rest.runpod.io/v2 unreachable (timed out after 4000ms or network error) — using v1 for now; will re-probe. Pin RUNPOD_REST_VERSION=v2 to force v2.`

Kept as an injected callback (defaulting to no-op) rather than a direct `console.*` so `backend.ts` stays pure/side-effect-free and unit tests stay silent. + tests: warns on throw/timeout and 5xx, does NOT warn on 404, and the no-op default doesn't crash. @jebenexer this closes the open thread from your #46 approval.

PR #47, opened 2026-06-24
[v2 4/4] Split + endpoint CRUD → v2 + new tools + spec-parity gate (BREAKING: default v2)

justinwlin:
**Part 4 of 4 — the v2 REST migration.** Closes the in-repo gaps from the [Agents Emergency Triage](https://www.notion.so/2fdff732fc348089bad5df8d0703b36b) the first three PRs left open: the per-resource source split, the rest of the v2 surface, **full v2 REST parity (endpoint CRUD migrated to v2)**, and a spec-parity drift gate.

> ⚠️ **Breaking change (major bump).** This PR (with #45/#46) makes **v2 the default** REST version *and* migrates Serverless endpoint CRUD to the v2 model. `create-endpoint`/`update-endpoint` move from the v1 **templateId** model to the v2 **inline-config** model (`imageName` + `gpuPoolIds` + `workers`/`scaling`; no `templateId`). An existing `create-endpoint` caller passing `{ templateId }` with no `RUNPOD_REST_VERSION` pin gets a clean `400` after upgrading — pin `RUNPOD_REST_VERSION=v1` to keep the legacy behavior. Changeset is **`major`**; README has a migration note.

## This PR

Commits, each independently reviewed to production-ready:

1. **Per-resource source split** — `src/tools.ts` (1577 lines) → a thin orchestrator + `src/tools/runtime.ts` (shared per-server runtime) + one module per resource (`pods`, `endpoints`, `jobs`, `templates`, `network-volumes`, `registries`, `catalog`). Zero behavior change: identical outbound requests (golden wire-locks), plus a frozen full-surface snapshot so a dropped registrar can't silently lose a tool.
2. **The rest of the v2 surface + full parity:**
   - **Tags** — `list/get/create/update/delete-tag` + `attach-tag`/`detach-tag`.
   - **Billing** — `get-billing` (scope all/pods/serverless/endpoints/networkvolumes/clusters; time window / `lastN`), records capped, `metadata` preserved.
   - **Endpoint workers** — `list-endpoint-workers` (`GET /v2/serverless/{id}/workers`), capped, `summary` preserved.
   - **Catalog by id** — `get-cpu-type`, `get-data-center` (round out `get-gpu-type`).
   - **Endpoint CRUD → v2** — `create/get/list/update/delete-endpoint` now route to `/v2/serverless` (inline `image`/`gpu.pools`/`workers`/`scaling`); the v1 templateId model is preserved when `RUNPOD_REST_VERSION=v1`. **Live-verified on prod v2** (create→get→update→delete, all resources torn down).
   - **Implemented but registration DISABLED** (commented out): `stream-pod-logs` (`GET /v2/pods/{id}/logs`, SSE) and `list-endpoint-releases` (`GET /v2/serverless/{id}/releases`). Their ops exist only on the **dev** v2 spec (47 ops) and **422 on prod** (45 ops), so exposing them would 422 in prod. Code + offline tests (SSE parser, mapper goldens) stay in the tree; re-enable when prod ships the ops. **Net registered surface: 50 tools.**
   - All v2-only tools: clean `501` on v1 (the `V2_ONLY` adapter stub). `capListResult` gained an `extra` arg so aggregate fields survive truncation.
3. **Spec-parity drift gate** — `tests/spec-parity.test.ts` walks every operation in the vendored v2 OpenAPI spec (`tests/fixtures/v2-openapi.yaml`, **dev superset = 47 ops**) and fails if any endpoint has no MCP tool (forward gate) or any tool is unaccounted-for (reverse gate), with stale-mapping / orphan-tool / parse-integrity guards. Hermetic; part of `pnpm test`. The two disabled-tool ops (`getPodLogs`, `listEndpointReleases`) are in `ALLOWLIST_UNMAPPED_OPS` with a documented reason. Optional live-shape check, skipped unless `MCP_LIVE_V2_KEY` is set. **Result: every spec op accounted for.**

`251 pass / 8 skipped (live) / 0 fail`, type-check + lint + build clean.

4. **#47 — Split, endpoint CRUD → v2, new tools, spec-parity gate** ← you are here

✅ done · 🔶 separate PR · ⬜ out of this repo / future cycle.

|---|---|---|
| Rest of the v2 REST surface (tags, billing, endpoint workers, catalog get-by-id) **+ endpoint CRUD on `/v2/serverless`** | ✅ | **#47** |
| Pagination + token-caps on **every** `list-*` | ✅ | base PR #32 + #45; extended to catalog/tags/billing/workers in #47 |
| MCP source split → `src/tools/<resource>.ts` + `_shared/` | ✅ | `_shared/` in #44; per-resource split in **#47** |
| Test suite walking every v2 spec endpoint → asserts a registered tool (CI merge gate) | ✅ | **#47** |
| Caller tracking | ✅ | merged PR #25; refactored in #44 |
| Hosted login / "Sign in with Runpod" OAuth | ✅ | merged PR #27 (PKCE enforcement tracked: DR-1398) |
| Steer agents to Flash | ⬜ | deferred |
| Cycle 3 (OpenAPI generator, GraphQL sunset, connector directory) | ⬜ | future |

> Note: the triage named "logs / artifacts / runs" as v2 surface. `getPodLogs` exists in the dev v2 spec and is **implemented** here (`stream-pod-logs`) but kept disabled until prod ships it; job *runs* are the serverless runtime API (`api.runpod.ai/v2`), covered by `run-endpoint`/`get-job-status`/`stream-job`/… (allowlisted as a different service).

Do not merge until v2 is live.


justinwlin:
## ✅ Full suite run against **prod v2** (`v2-rest.runpod.io/v2`) — 2026-06-26

v2 prod is now cut over and authenticates with a prod key (allowlist-gated). Drove the entire `test.md` suite (A–I, all CRUD) end-to-end through the MCP over stdio. **All pass, every resource torn down, no `mcptest-*` leaks** on v1 or v2.

| Section | Result |
|---|---|
| A — catalog reads + get-by-id (gpu/cpu/dc) | ✅ |
| B — templates full CRUD | ✅ (incl. `containerRegistryAuthId` is a separate PR #48 on `main`) |
| C — container registry auth CRUD | ✅ |
| H — tags CRUD + attach/detach to a live pod | ✅ |
| I — billing (all + scoped) | ✅ |
| D — network volume CRUD (billable storage) | ✅ created/grown/deleted @AP-JP-1 |
| E — pod lifecycle (RTX 3070 community @ $0.13/hr) | ✅ create→get→stop→start→restart→delete |
| list-endpoint-workers | ✅ against a live endpoint |
| F — v1↔v2 parity (reads + template CRUD on v1) | ✅ same shapes |
| G1/G2 — v2-only tools under v1 | ✅ clean 501 notices (list-cpu-types, get-gpu/cpu/dc-type, restart-pod, list-tags, get-billing) |
| G3 — CPU pod on v2 → routes to v1 | ✅ `_servedBy:"v1"` |
| G4/G5 — ambiguous / contradictory create | ✅ clean 400, no request fired |

Routing confirmed correct: pods/templates/network-volumes(`/network-volumes`)/registries(`/registries`)/tags/billing/catalog → v2; **endpoints → v1** (v2 has no `/v2/endpoints` yet); **CPU pods → v1**.

Also included in this push: fixed `scripts/smoke.ts` `summarizeGpuTypes` to read v2 field names (`name`/`memory`) — it previously only read v1 names so v2 smoke output showed a bare `id`. Server behaviour was always correct; display-only fix. Full results table + sign-off are in `test.md`.

justinwlin:
## ✅ Full suite re-run against the **deployed HTTP server** (v2) — 2026-06-26

Validated the same A–I suite against a Vercel deploy configured with `RUNPOD_REST_VERSION=v2` (`runpod-ejf4ucryo-...vercel.app`), driving it over the **Streamable HTTP transport** (key as per-request `Authorization: Bearer`, not env). This complements the earlier stdio run — same tool surface, same results, over the real hosted path.

| Section | Result |
|---|---|
| handshake / auth / `tools/list` | ✅ 50 tools |
| A — catalog reads + get-by-id (gpu/cpu/dc) | ✅ |
| B — templates full CRUD | ✅ |
| C — container registry auth CRUD | ✅ |
| H — tags CRUD + attach/detach to a live pod | ✅ |
| I — billing (all + scoped) | ✅ |
| D — network volume CRUD | ✅ created/grown/deleted |
| E — pod lifecycle (RTX 3090 community @ $0.22/hr) | ✅ create→get→stop→start→restart→delete |
| list-endpoint-workers | ✅ |
| G3 — CPU pod → routes to v1 | ✅ `_servedBy:"v1"` |
| G4 / G5 — ambiguous / contradictory create | ✅ clean 400, **no pod created** |

**Teardown:** every created resource deleted; final `list-pods` = 0; no `mcptest-*` leaks (pods/templates/volumes/registries/tags all clean).

Notes:
- Version on the hosted transport is a **server-side env decision** (`RUNPOD_REST_VERSION` on the deploy), not a per-request client option — a default-v1 preview serves v1 over HTTP until the env is set to v2.
- The ambiguous/contradictory-create guards (G4/G5: clean 400, no request) are **v2-only**. On a v1 deploy `create-pod` is passthrough and v1 will happily spin up a pod from an underspecified request — worth keeping in mind for any v1 hosted deploy.

src/tools/tags.ts
```diff
@@ -0,0 +1,214 @@
...
+        .string()
+        .describe('ID of the resource to attach the tag to'),
+    },
+    { title: 'Attach tag to resource', ...WRITE, idempotentHint: true },
+    async (params) => {
+      const backend = backendFor('tags');
+      if (backend.version === 'v1')
+        return jsonReply(v2OnlyNotice('attach-tag'));
+      const result = await callRestUrl(
+        `${backend.base}/tags/${params.tagId}/resources/${params.resourceType}/${params.resourceId}`,
```
LarveyOfficial: 🟡 Medium: tagId and resourceId are arbitrary caller strings interpolated into the URL path unencoded. resourceType is enum-constrained (safe), but an id containing / ? # or % mis-routes the request or breaks path segmentation (an id with ? turns the rest into a query string). Same-tenant so it's not a cross-tenant security hole, but it's a correctness bug — wrap both in encodeURIComponent(). (Same unencoded pattern in the detach handler below — fix both.)

src/tools/billing.ts
```diff
@@ -0,0 +1,101 @@
...
+          status: 501,
+        });
+      }
+
+      const subPath = BILLING_SCOPE_PATH[params.scope ?? 'all'];
+      const query = new URLSearchParams();
+      if (params.startTime) query.set('startTime', params.startTime);
+      if (params.endTime) query.set('endTime', params.endTime);
+      if (params.bucketSize) query.set('bucketSize', params.bucketSize);
+      if (params.lastN !== undefined) query.set('lastN', String(params.lastN));
```
LarveyOfficial: 🟡 Medium: the schema descriptions say startTime/endTime are "mutually exclusive with lastN," but the handler sets all three if present, forwarding a contradictory request whose behavior depends on undefined upstream precedence. Add a Zod .refine() rejecting the combination, or drop lastN when a window is supplied. Low blast radius (read-only), hence medium.

src/pagination.ts
```diff
@@ -61,9 +61,15 @@ export function encodeCursorOffset(offset: number): string {
...
+//
+// `extra` merges sibling fields alongside `items`/`pagination` (e.g. billing
+// `metadata`, an endpoint-workers `summary`) so a capped list response can still
+// carry the small aggregate fields that accompany the array. Reserved keys
+// (`items`, `pagination`) in `extra` are ignored so they can't shadow the cap.
 export function capListResult(
   result: unknown,
-  options: { limit?: number; cursor?: string }
+  options: { limit?: number; cursor?: string },
+  extra?: Record<string, unknown>
```
LarveyOfficial, line 72: 🔵 Low: when the result isn't an array, capListResult returns early and drops this new `extra` (metadata/summary/endpointVersion). In practice unwrapList always yields an array so it doesn't bite today, but if unwrap ever returns records under an unexpected key the agent gets an empty page *and* loses the metadata extra was meant to preserve. Consider merging extra in the early-return branch too.

tests/spec-parity.test.ts
```diff
@@ -0,0 +1,301 @@
...
+      break;
+    }
+  }
+
+  const ops: SpecOp[] = [];
+  let currentPath: string | null = null;
+  for (const line of lines.slice(start + 1, end)) {
+    const pathMatch = /^ {2}(\/[^:]+):\s*$/.exec(line);
+    const methodMatch = /^ {4}(get|post|put|patch|delete):\s*$/.exec(line);
+    const oidMatch = /^ {6}operationId:\s*(\S+)/.exec(line);
```
LarveyOfficial, line 49: 🔵 Low: the hand-rolled parser depends on a 2/4/6-space indentation contract and will mis-parse if the vendored spec is ever reformatted. It's well-mitigated by the "parsed EVERY operationId" cross-check guard (a silent undercount fails loudly) — flagging only so reviewers know that guard is load-bearing: please don't remove it.

scripts/serve-http.local.ts
```diff
@@ -0,0 +1,26 @@
+// LOCAL DEV ONLY (uncommitted) — boots the MCP server over HTTP so an agent can
```
LarveyOfficial: 🔵 Nit: the banner says the file is uncommitted, but the PR commits it. Harmless, but the comment is now misleading — reword to "local dev helper".

LarveyOfficial (commented):
Automated review pass — read the full diff (plus the truncated tail of the new modules from the PR head). Clean extraction with wiring centralized in createToolRuntime; no critical issues. Two mediums inline (unencoded user input in tag attach/detach URLs; unenforced billing start/end-vs-lastN exclusivity), rest are polish. Verified correct: unwrapList safely coerces null/non-object → []; v2-only resources (tags/workers/billing) resolve to a v1 stub that triggers a clean 501; capListResult strips reserved keys from extra; the auto-probe is forced to v1 on HTTP transport to prevent cross-user verdict leakage; spec-parity asserts both directions.

justinwlin:
## Addressed the review's top-priority findings — commit d344305

The four 🚩 items from the review breakdown are resolved here (the per-resource split consolidated these files on this PR, so the fix lands at the stack tip; it flows down to the merged result and was rebased into #49's preview too).

| Finding | Verdict after investigation | Fix |
|---|---|---|
| **#3 data centers** | **Real bug — worse than flagged.** The mapper emitted a singular `dataCenter` key that doesn't exist on the v2 `CreatePodRequest` (the field is **`dataCenterIds`, an array**, spec L943). So DC placement was dropped entirely on v2 pod create, not just truncated. | `mapPodCreateToV2` now emits `dataCenterIds` (full array, dropped when empty). Fixture + unit/handler tests updated to the correct shape. |
| **#3 multi-GPU** | Valid (silent narrowing). v2 `GpuConfig` is genuinely singular, so `[0]` is forced — but it was silent. | create-pod now echoes a `_warning` when >1 `gpuTypeId` is supplied, naming the ignored ids. |
| **#2 template category** | Valid. `create-template` never exposed `category`, so the mapper always defaulted to `NVIDIA` — CPU/AMD templates mis-categorized, no override. | Added optional `category: CPU\|NVIDIA\|AMD` to the tool schema (flows through the existing mapper default). |
| **#1 CPU fallback error handling** | Overstated (the catch only converts 501, which v1 won't emit; v1 `mapCreate` is identity, and we'd already proven CPU create works live). Tidied for consistency anyway. | Wrapped the CPU v2→v1 fallback in try/catch → clean error reply instead of a raw throw. |

Also: fixed a now-stale mapper header comment.

**Not changed — #4 (probe `AbortSignal`):** we're on `node-fetch@3`, which honors WHATWG `AbortSignal` (incl. `AbortSignal.timeout`); the "may be a no-op" caveat is a node-fetch@2 issue. The guard works. Only gap was test coverage of real cancellation — noted, optional.

Verified: `tsc` clean, `eslint` clean, `pnpm test` **214 pass / 0 fail** (added handler tests for the multi-GPU warning and the CPU-fallback error path). An independent review subagent confirmed all four fixes correct against the vendored v2 spec before this was committed.

src/tools/tags.ts
justinwlin: Fixed in 2d8685d — both attach-tag and detach-tag now `encodeURIComponent` the free-text `tagId`/`resourceId` path segments (resourceType is enum-constrained, left raw). + handler test asserting `a/b?c`→`a%2Fb%3Fc`, `p#1`→`p%231`.

src/tools/billing.ts
justinwlin: Fixed in 2d8685d — `get-billing` now drops `lastN` when a `startTime`/`endTime` window is supplied, so it never forwards a contradictory request. + handler test.

src/pagination.ts
justinwlin: Noted — doesn't bite today (`unwrapList` always yields an array), so it's a latent robustness gap; merging `extra` into the early-return branch is a reasonable follow-up.

tests/spec-parity.test.ts
justinwlin: Noted — the load-bearing 'parsed EVERY operationId' cross-check guard stays; we won't remove it. Agreed it's what makes the indentation-based parser safe.

scripts/serve-http.local.ts
justinwlin: Fixed in 2d8685d — reworded the banner to 'LOCAL DEV HELPER' (no longer claims it's uncommitted).

justinwlin:
## Defaulted the local REST version to v2 — commit 2658e43

Per request, flipped the unset `RUNPOD_REST_VERSION` default from **v1 → v2** in `resolveVersion`, and made the test suite independent of the default so changing it no longer dictates pass/fail.

**Behavior:**
- With nothing set, control-plane resources now use **v2**. Set `RUNPOD_REST_VERSION=v1` to pin the old behavior.
- Unchanged: `jobs`/`endpoints` still force v1 (no v2 home yet), CPU pods still served by v1, the `auto`→v1-on-HTTP guard, and bogus values still fall back to v1.

**Tests — no longer depend on the ambient default** (this was the ~58-test breakage when the default flips):
- `handlers.test.ts`: the global `beforeEach` now pins `RUNPOD_REST_VERSION=v1`, so the "v1 unchanged" goldens + the 501-on-v1 cases assert v1 *explicitly*; v2 cases still opt in via `withV2(...)`.
- `backend.test.ts`: the `resolveBackend (v1)` shared env + the two base-override tests pin v1 explicitly; `unset defaults to v1` → **`unset defaults to v2`**, now the single source of truth for the default.

**Docs:** README version section (intro, table Default column, notes), the `serve-http.local.ts`/`test.md` banner (`v2(default)`), and a changeset all updated.

Verified by an independent review subagent: default is genuinely v2, **no version-sensitive test relies on the ambient default** anymore, docs consistent, staged set clean (7 files, zero deletions). Gate: tsc ✓, eslint ✓, `pnpm test` **216 pass / 0 fail / 8 skipped**, build ✓. Rebased onto #49 so its preview carries it too.

justinwlin:
## v2 full-parity addendum — last 3 gaps closed

Pushed `c9a68aa`. The MCP now uses v2 everywhere v2 exists. Diffed the full v2 spec (47 ops) against the tool surface — these were the only remaining gaps:

### 1. Endpoint CRUD migrated to v2 `/v2/serverless`
v2 endpoints use an **inline** config (`image` + `gpu.pools` + `workers` + `scaling`) — no `templateId`. Routing is **version-aware**:
- **v2** (default): `create-endpoint` takes `imageName` + `gpuPoolIds` (GPU **pool** names from `list-gpu-types`, e.g. `AMPERE_80`), plus optional workers/scaling/disk/env/registry/networkVolumes/timeout/flashboot. Guards the required fields with a clean `400` before any request.
- **v1**: the legacy `templateId` model is preserved unchanged.
- `list`/`get`/`update`/`delete` route through the adapter on both versions (v1-only `includeTemplate`/`includeWorkers` query params are dropped on v2).

### 2. `list-endpoint-releases` (new tool)
`GET /v2/serverless/{id}/releases` — v2-only (clean 501 on v1). Caps the `releases` array, preserves `rollout` + `endpointVersion`.

### 3. `stream-pod-logs` (new tool)
Bounded SSE read of `GET /v2/pods/{id}/logs` (v2-only, 501 on v1). Reads up to `maxWaitMs` (default 5s) / 256 KB and returns the collected `{source,line,ts}` lines. Pure `parsePodLogSse` frame parser + a time/byte-bounded `streamSse` runtime reader (collects Buffers, decodes once — no multibyte-split corruption).

### Verification
- Spec re-vendored from dev (45→47 ops); **spec-parity gate green** (every op mapped). Surface **50→52** tools. Suite **219→253** tests, all green; `tsc` + `eslint` clean.
- **Endpoint CRUD live-verified on prod v2** (`v2-rest.runpod.io/v2`): create → get → update → `list-endpoint-releases` → `list-endpoint-workers` → delete. All `mcptest-*` resources torn down; pre-existing endpoints untouched.
- `getPodLogs` + `listEndpointReleases` are **dev-ahead-of-prod** (present in dev's 47-op spec, not yet on prod's 45 — prod returns `422 path not found`), so they're verified offline (SSE-parser unit tests, v2 wire-locks, v1→501) and live on dev only. Noted in `test.md` (§J/§K + results).
- Adversarial subagent review: no blocking issues across routing / mappers / guards / SSE parsing / stream bounding.

`test.md` updated with the J (endpoint v2 CRUD) and K (pod logs) suites + filled prod results.

justinwlin:
### Follow-up `e98c965` — gated the two dev-only tools off

`stream-pod-logs` and `list-endpoint-releases` are now **registered-disabled** (their `server.tool(...)` blocks are commented out). Their ops — `GET /v2/pods/{id}/logs` and `GET /v2/serverless/{id}/releases` — exist only on the **dev** v2 deployment (`v2-rest.runpod.dev`, 47-op spec); **prod** (`v2-rest.runpod.io`, 45-op spec) returns `422 "path not found"`. Exposing tools that 422 in prod isn't useful, so they stay off until prod ships the ops.

- Implementations + offline tests (SSE parser, mappers) remain in the tree.
- The two dev-only ops moved to `ALLOWLIST_UNMAPPED_OPS`, so the parity gate stays green against the dev superset spec.
- Registered surface: **52 → 50**. Suite still green (248 tests). Re-enable = uncomment the blocks + revert the test entries, once prod has the ops.

**Endpoint CRUD on `/v2/serverless` is unaffected** — that op IS on prod and stays live-verified.

.changeset/default-rest-v2.md
jebenexer, line 1: This should be a **major** bump, not minor.

The default switching from v1 to v2 silently breaks any existing `create-endpoint` caller who has no `RUNPOD_REST_VERSION` pin. V1 `create-endpoint` is `templateId`-based; v2 is inline container-config — a call with `{ templateId: '...' }` that worked on v1 now returns a clean `400` after upgrading, without the caller changing anything. A `minor` bump signals backward compatibility; this isn't.

Please update the changeset to `major` and add a migration note in the README that covers the `create-endpoint` schema change specifically (the env-var escape valve is documented, but the breaking field-level difference is not).

src/tools/endpoints.ts
jebenexer, line 1: **Two issues in the `create-endpoint` handler:**

**1. Missing `name` guard — raw 422 instead of clean 400**

The PR description says "the handler guards required fields so the caller gets a clean 400 rather than a raw 422." That's only half true. `imageName` and `gpuPoolIds` are guarded, but `name` is not. The v2 spec marks `name` as required on `CreateEndpointRequest`. A caller who omits it passes both guards, the mapper emits a body without `name`, and the API returns an opaque 422. The Zod schema marks it `z.string().optional()` so there's no schema-level catch either. Add alongside the existing guards:
```ts
if (!params.name) {
  return jsonReply({ error: 'create-endpoint on v2 requires a name.', status: 400 });
}
```

**2. v1 passthrough sends all params including v2-only fields**

The v1 path sends the full Zod-validated `params` object to the v1 API, including v2-only fields like `gpuPoolIds`, `flashboot`, `networkVolumeIds`, `executionTimeoutMs`. V1 likely ignores unknown fields today, but it's imprecise and will silently send garbage if v1 ever strict-validates or a field name collides. Should filter to v1-relevant fields (`templateId`, `name`, `computeType`, `gpuTypeIds`, `gpuCount`, `workersMin`, `workersMax`, `dataCenterIds`) before the call.

src/tools/runtime.ts
jebenexer, line 1: **Two issues in this file:**

**1. `graphqlRequest` bypasses the injected `deps.fetch` seam**

`createToolRuntime` captures `httpFetch = deps.fetch ?? defaultFetch` and threads it into `v1Client`, `serverlessClient`, and `restClient`. But `graphqlRequest` calls the module-level imported `fetch` directly — it is never given `httpFetch`. This means:

- Injecting a fake fetch in tests does **not** intercept GraphQL calls.
- The claim in `handlers.test.ts` that everything is "all offline — no network, no real fetch" is broken for the v1 paths of `list-gpu-types` and `list-data-centers`.

Fix: pass `httpFetch` into `graphqlRequest` as a parameter and capture it in the `ToolRuntime` object alongside the other clients.

**2. Module-level URL constants are frozen at import time**

`API_BASE_URL`, `SERVERLESS_API_BASE_URL`, and `GRAPHQL_URL` are evaluated once when the module is first imported. `callRestUrl` reads env live via `backendFor`; the CPU-pod v1 fallback uses `restV1Base(env)` (live). But `runpodRequest` and `serverlessRequest` use the frozen module-level constants. This is an inconsistent lifecycle — if env is changed between tests without a module reload, the wrong URL is silently used. Consolidate to the `Env` abstraction pattern already used by the rest of the adapter.

scripts/serve-http.local.ts
jebenexer, line 1: `server.listen(PORT)` without a host argument defaults to `0.0.0.0` — all interfaces. The script is described as a "LOCAL DEV HELPER" and the log output says `http://localhost:${PORT}`, but the process accepts connections from any IP. On a cloud VM or CI runner this is reachable from outside the host. Fix: `server.listen(PORT, '127.0.0.1', callback)`.

jebenexer (changes requested):
Three must-fix items plus a couple of minors. The semver bump is the most user-visible — any existing `create-endpoint` integration without an explicit `RUNPOD_REST_VERSION` pin will silently break on upgrade. The `graphqlRequest` seam gap means the v1 GPU/DC golden tests aren't actually offline. The `name` guard omission contradicts the stated design goal of clean 400s.

justinwlin:
Thanks @jebenexer — all addressed on the tip (`v2/group-4-source-split`, commit `29361d0`):

**Blockers**
- **Changeset → `major`** + README migration note. Corrected the stale "endpoints always v1" line and documented the `create-endpoint` v1→v2 break (templateId model → inline `imageName`+`gpuPoolIds`; a `{templateId}` call now returns a clean 400 unless `RUNPOD_REST_VERSION=v1`).
- **`create-endpoint` missing `name` guard** — added (spec-required); now a clean 400 instead of a raw 422. + test.
- **`graphqlRequest` bypassed the injected fetch** — now routed through `httpFetch`, so the v1 catalog (`list-gpu-types`/`list-data-centers`) goldens are genuinely offline. Added a test that captures the GraphQL call via the fake fetch.

**Mediums**
- **v1 `create-endpoint` passthrough** now filters to v1-only fields (drops `flashboot`/`gpuPoolIds`/`networkVolumeIds`/… from the v1 body). + test.
- **Module-level URL consts** (`runpodRequest`/`serverlessRequest`/graphql) now resolve base URLs **live** per call (via `restV1Base`/`serverlessBase`/`publicGraphqlBase`), matching `callRestUrl` — no more frozen-at-import inconsistency.

251 tests green, tsc/eslint/build clean.

.changeset/default-rest-v2.md
justinwlin: Fixed in `29361d0` — changeset bumped to **`major`**, and a README migration note added (corrected the stale "endpoints always v1" line too). The note documents the `create-endpoint` v1→v2 break: templateId model → inline `imageName`+`gpuPoolIds`; a `{templateId}` call now returns a clean `400` unless `RUNPOD_REST_VERSION=v1`.

src/tools/endpoints.ts
justinwlin: Both fixed in `29361d0`:

1. **Missing `name` guard** — added (v2 spec marks it required); now a clean `400` before any request instead of a raw 422. + test.
2. **v1 passthrough** — now filters to v1-relevant fields (`templateId`, `name`, `computeType`, `gpuTypeIds`, `gpuCount`, `workersMin`, `workersMax`, `dataCenterIds`); v2-only keys (`flashboot`/`gpuPoolIds`/`networkVolumeIds`/…) no longer reach the v1 API. + test asserting they're dropped.

src/tools/runtime.ts
justinwlin: Both fixed in `29361d0`:

1. **`graphqlRequest` seam** — now routed through the injected `httpFetch`, not module-level node-fetch. The v1 catalog (`list-gpu-types`/`list-data-centers`) goldens are now genuinely offline; added a test that captures the GraphQL POST via the fake fetch.
2. **Frozen module URLs** — `runpodRequest`/`serverlessRequest`/graphql now resolve their base URLs **live** per call (`restV1Base`/`serverlessBase`/`publicGraphqlBase`), matching `callRestUrl`. No more import-time freeze inconsistency.

scripts/serve-http.local.ts
justinwlin: Fixed in `29361d0` — now `server.listen(PORT, '127.0.0.1', …)` so the dev helper binds loopback only, not `0.0.0.0`. Added a comment explaining why (it holds per-request keys and shouldn't be reachable off-host on a VM/CI runner).

jebenexer (approved):
All items from the follow-up review verified fixed in 29361d0. Specifically confirmed in the diff:\n\n- **Semver** → `major` bump with a clear migration note in the changeset and README.\n- **`create-endpoint` name guard** → `!params.name` guard added before `imageName`/`gpuPoolIds` checks; v1 passthrough now filters to v1-relevant fields only (v2-only keys dropped, test added).\n- **`graphqlRequest` seam** → now takes `fetchImpl`/`url` params, wired through the injected `httpFetch`; v1 catalog goldens are genuinely offline. Test added capturing the GraphQL POST via the fake.\n- **Frozen URL constants** → removed; v1/serverless/graphql base URLs now resolved live per call via `restV1Base`/`serverlessBase`/`publicGraphqlBase`, matching `callRestUrl`.\n- **`serve-http.local.ts`** → binds `127.0.0.1` with explanatory comment.\n\n251 tests green per the commit message. Stack is ready to land.

PR #49, opened 2026-06-26
[v2 5/5] Advertise S256 PKCE so Claude Code can start the OAuth flow

justinwlin:
Stacked on #47 (`v2/group-4-source-split`).

**Why this PR:** Claude Code (and other RFC 8414 OAuth clients) refuse to start the sign-in flow unless the server advertises `S256` PKCE in its discovery metadata. Ours didn't, so "Sign in with Runpod" failed before the browser ever opened. This one-line metadata fix unblocks them. **Enforcing** the challenge is a tracked follow-up — see below.

4. **#47 — Split, endpoint CRUD → v2, new tools, spec-parity gate**
5. **#49 — Advertise S256 PKCE so OAuth clients (Claude Code) can start the flow** (tip) ← you are here

## Problem
Clients read `/.well-known/oauth-authorization-server` during discovery and **refuse to start the flow** unless `code_challenge_methods_supported` lists `S256`. Our metadata omitted the field, so the flow errored before the browser opened.

## Change
One line in `handleAuthorizationServerMetadata` (`api/index.ts`):
```json
"code_challenge_methods_supported": ["S256"]
```

## Important: advertise-only for now (deliberate stopgap)
The `code_challenge` is **not yet verified** server-side. The MCP server is a stateless Vercel function with nowhere to bind the challenge across the `/authorize` → `/token` boundary (and `/authorize` doesn't even mint the code — the console does, from the flash `requestId`). So this PR makes the discovery check pass; it does not yet enforce PKCE.

This is acceptable as a stopgap because the flow's real protection does not depend on PKCE — all authorization is flash/console-side:
- the authorization code is the flash request id — **single-use, ~10-min TTL, requires a human to approve in the console**, and
- the code is only ever delivered to an **allowlisted redirect_uri** (`claude.ai` / `claude.com` / loopback), re-checked at both `/authorize` and `/token`.

The one path where verified PKCE would add real protection is **loopback** (native-app interception); the compensating controls blunt it but don't fully close it.

## Follow-up — tracked in **[DR-1398](https://linear.app/runpod/issue/DR-1398)** (High priority)
Enforce S256 by storing `code_challenge` across the boundary and verifying `base64url(sha256(code_verifier))` at `/token` (reject `invalid_grant` on mismatch), plus **require PKCE for loopback** redirects. Two implementation options in the ticket: (A) persist on the `FlashAuthRequest` row (`model/src/flash/authRequests.ts` + flash GraphQL) — recommended; (B) Vercel KV keyed by `requestId` — MCP-only, no backend change.

**We will not advertise "PKCE supported" as a security feature publicly until enforcement lands** — the discovery metadata stays (it's a compatibility shim), but no changelog/marketing claim until DR-1398 is done.

## Test
Deploy a preview, connect Claude Code with no key, confirm "Sign in with Runpod" now gets past discovery and completes.

### DIFF
--- a/api/index.ts
+++ b/api/index.ts
@@ -235,9 +235,15 @@ function handleAuthorizationServerMetadata(
     response_types_supported: ['code'],
     grant_types_supported: ['authorization_code'],
     token_endpoint_auth_methods_supported: ['none'],
-    // Note: PKCE is intentionally not advertised because it is not enforced
-    // server-side (the flash flow has no place to bind the code_challenge).
-    // Security rests on the redirect_uri allowlist + single-use flash approval.
+    // Advertise S256 PKCE so OAuth clients that require it (e.g. Claude Code,
+    // which refuses to start the flow without this field per RFC 8414/8252)
+    // will proceed. NOTE: the code_challenge is not yet verified server-side —
+    // the stateless function has nowhere to bind it across /authorize→/token.
+    // Until enforcement lands (persist code_challenge on the flash auth request
+    // and check the verifier at /token), the flow's real protection remains the
+    // redirect_uri allowlist + single-use, short-lived, human-approved flash
+    // request. Advertising S256 here unblocks clients without weakening that.
+    code_challenge_methods_supported: ['S256'],
   });
 }
 


api/index.ts
```diff
@@ -235,9 +235,15 @@ function handleAuthorizationServerMetadata(
...
-    // Security rests on the redirect_uri allowlist + single-use flash approval.
+    // Advertise S256 PKCE so OAuth clients that require it (e.g. Claude Code,
+    // which refuses to start the flow without this field per RFC 8414/8252)
+    // will proceed. NOTE: the code_challenge is not yet verified server-side —
+    // the stateless function has nowhere to bind it across /authorize→/token.
+    // Until enforcement lands (persist code_challenge on the flash auth request
+    // and check the verifier at /token), the flow's real protection remains the
+    // redirect_uri allowlist + single-use, short-lived, human-approved flash
+    // request. Advertising S256 here unblocks clients without weakening that.
+    code_challenge_methods_supported: ['S256'],
```
LarveyOfficial, line 246: ✅ Correct and safe as a stopgap. Right field/value per RFC 8414, additive, and nothing depends on `code_challenge` server-side — unblocks Claude Code with no behavior change.

The advertise-without-enforce tradeoff is fine for the hosted `claude.ai`/`claude.com` redirect (TLS + exact-match host makes code interception impractical). The path where the missing PKCE verification actually matters is **loopback** (`isRedirectUriAllowed` allows `http://` on any port) — that's the native-app interception scenario PKCE was built for. Single-use + human-approval + TTL narrow the race window but don't fully close it there, so the enforcement follow-up matters most for loopback clients. Honest comment; no change requested here.

LarveyOfficial (commented):
Sanity-check pass on the OAuth change — read the full /authorize + /token flow, not just the diff. **The change is correct and safe to ship as a stopgap.** `code_challenge_methods_supported: ['S256']` is the right RFC 8414 field/value, it's purely additive, and nothing in the flow reads or depends on `code_challenge` (only logged as `hasCodeChallenge`), so it unblocks Claude Code's discovery check with no behavior change or breakage.

On the advertise-but-don't-enforce tradeoff: PKCE protects against authorization-code interception, and the compensating controls here genuinely blunt that — the code is single-use/consumed atomically at /token, only yields a key after a human approves the request in the console, and is only delivered to an allowlisted redirect_uri (re-checked at both endpoints). For the hosted `claude.ai`/`claude.com` clients that reasoning is solid (TLS to an exact-match host; interception isn't practical).

One nuance worth calling out: the path where this actually matters is **loopback** (`http://` any port). Loopback/native redirects are exactly the case PKCE was designed for — a local process on a shared machine can race for the code. The controls narrow that window (attacker still needs the user to approve *that* request, within the TTL, and to win the single-use redemption race) but don't fully close it the way a verified challenge would. So residual risk is concentrated on the loopback path — which is precisely what the tracked enforcement follow-up addresses. Not a blocker; just make sure that follow-up doesn't rot.

Two backend-side assumptions the argument leans on that I couldn't verify from this repo (they live in `model/src/flash/authRequests.ts`) and would just confirm: the **~10-min TTL** and that **approval is bound to the specific request id**. The /token code corroborates single-use; the TTL + per-request approval are the load-bearing parts.

api/index.ts
justinwlin: Noted, and agreed on the analysis — the enforcement follow-up matters most for the loopback redirect (the native-app interception case PKCE was built for; the hosted claude.ai/claude.com path is TLS + exact-host). It's tracked in this PR's description (persist `code_challenge` on the FlashAuthRequest row + verify the verifier at /token) so it won't rot. No change requested here — thanks for the careful read.

api/index.ts
jebenexer, line 1: The PR description acknowledges this honestly, but the failure mode for strict PKCE clients is worth stating precisely: **pre-PR**, a strict OAuth client that requires `S256` in discovery metadata refused to start the flow at discovery time (a clean, early error). **Post-PR**, it starts the flow, sends `code_challenge`, and then fails when it validates that the server echoed PKCE parameters in the token response — because the server silently ignores `code_challenge`. This is a compatibility regression: clients that previously handled a graceful "S256 not advertised" refusal now fail at a later and potentially less graceful point.

The flow's real protection (redirect URI allowlist + single-use human-approved request) is unchanged, so this isn't a security regression in terms of what an attacker can do. But please track the enforcement work (persist `code_challenge` on the flash auth request, verify `base64url(sha256(code_verifier))` at `/token`) as a blocker before any public announcement of PKCE support — advertising S256 without enforcing it creates a misleading security signal for integrators reading the discovery doc.

jebenexer (commented):
One item: the specific failure mode for strict PKCE clients differs from what the PR description implies. The security properties are unchanged, but the compatibility behavior regresses for clients that handled a clean pre-discovery refusal and now encounter a later failure instead.

justinwlin:
Thanks @jebenexer / @LarveyOfficial — agreed on the precise framing (strict clients now fail later rather than at discovery, and the loopback path is where verified PKCE would actually matter; security properties are unchanged).

**Decision: ship the advertise-only metadata as a stopgap to unblock Claude, enforce as a tracked follow-up.** Rationale: the MCP server is a stateless relay — all real authorization is flash/console-side (redirect_uri allowlist re-checked at both endpoints, single-use code consumed atomically, short TTL, human approval). The S256 advertisement only satisfies clients' discovery check.

Per both of your conditions:
1. **Tracked:** **DR-1398** — https://linear.app/runpod/issue/DR-1398 (High priority). It captures both implementation options (A: persist `code_challenge` on the flash request via `createFlashAuthRequest`/`flashAuthRequestStatus` — recommended; B: Vercel KV keyed by requestId), the `base64url(sha256(verifier))` verification, and a **require-PKCE-for-loopback** policy to close the residual-risk path.
2. **No public "PKCE supported" claim** until enforcement lands — the discovery metadata stays, but it won't be announced as a security feature.

No code change here beyond the existing advertisement.

api/index.ts
justinwlin: Agreed on the precise framing (strict clients now fail later rather than at discovery; loopback is where verified PKCE matters; security properties unchanged). 

Decision: keep the advertise-only metadata as a deliberate stopgap to unblock Claude (all real auth is flash/console-side), and enforce as a tracked follow-up — **[DR-1398](https://linear.app/runpod/issue/DR-1398)** (High). It captures both options (A: persist `code_challenge` on the flash request; B: Vercel KV keyed by requestId), the `base64url(sha256(verifier))` check, and a **require-PKCE-for-loopback** policy. We won't advertise PKCE as a security feature publicly until enforcement lands. PR description updated to link the ticket.

PR #52, opened 2026-07-06
feat: enable pod + serverless worker log streaming (v2)

justinwlin:
## What this does

Adds live log streaming for v2, exposed as two MCP tools:

- **`stream-pod-logs`** — `GET /v2/pods/{id}/logs`. Was already implemented but its registration was commented out while the endpoint was dev-only. It's live on prod now, so this enables it.
- **`stream-worker-logs`** (new) — `GET /v2/serverless/{id}/workers/{workerId}/logs`, for a single serverless worker. Get the `workerId` from `list-endpoint-workers`.

Both read a **bounded snapshot** of the Server-Sent-Events stream (an MCP tool returns once, not an open stream): they read until `maxWaitMs` elapses or a 256 KB cap is hit, then return the parsed frames.

## How to use

Params (shared by both tools): `source` (`container` | `system` | `both`, default `both`), `tail`, `since`, `maxWaitMs`. Both are v2-only — on v1 they return a clean `501` notice without making a request.

Result shape:

```json
{
  "items": [
    { "source": "system", "line": "create 20GB volume", "ts": "2026-07-06T22:04:13Z" }
  ],
  "count": 1,
  "truncated": false
}
```

## Design

The two tools are the same feature on two resources, so the shared logic lives in **`src/tools/logs.ts`**:

- `parseLogSse` — parses SSE `data: {source,line,ts}` frames (handles multi-line `data:`, CRLF, `id:`/`event:`/comment lines, non-JSON payloads).
- `collectLogSnapshot` — builds the query string and does the bounded read.
- `streamLogsReply` — the whole tool handler (backend resolve → v2 gate → snapshot → error mapping).

Each tool registration is then just a name + a URL builder.

## Included fixes

- **Query bug:** `source: 'both'` now omits the `source` param (the wire enum is only `container|system`; the endpoint returns both when it's absent) instead of sending an invalid `source=both`.
- **Reader bug (found in review):** the bounded SSE reader could throw an `AbortError` if the `maxWaitMs` timeout fired *before* response headers arrived (low `maxWaitMs` + slow backfill), erroring the tool instead of returning a partial snapshot. Fixed so a timeout during connect or body-read is treated as a normal end.

## Testing

- **Offline:** 253 pass / 0 fail. `parseLogSse` unit tests + handler tests that drive both tools through an injected reader and lock the URL/query building and parsed output. Refreshed the vendored v2 OpenAPI fixture and mapped both ops in the spec-parity gate.
- **Type-check / eslint / prettier:** clean.
- **Live (prod):** streamed a real pod (`stream-pod-logs`) — `both` = 43 lines (6 system + 37 container), `source=system` = 6, `source=container` = 37, correct `{source,line,ts}` frames. Confirmed the `501` v1 guard for both tools.

## Notes

`list-endpoint-releases` (`GET /v2/serverless/{id}/releases`) is now live on prod too, but is intentionally left disabled to keep this PR scoped to log streaming.


justinwlin:
## Review follow-up

Addressed the requested change plus the two minors (commit c22e0c6).

**Requested change — the real SSE reader had zero test coverage.** Every handler test injected a fake `streamSse`, so the production reader (timeout `AbortController`, connect-vs-body abort handling, 256 KB cap, once-at-the-end decode) never actually ran — a regression there would keep CI green.

Fix: extracted the production read loop into an exported `readSseSnapshot(fetchImpl, url, headers, opts)` in `logs.ts` (production wires it to node-fetch in `createToolRuntime`; behavior unchanged). New `tests/logs.test.ts` drives the *real* reader against a mock `Response`:
- body over the cap → `truncated: true` and the stream is aborted
- `AbortError` thrown while **connecting** → partial snapshot, no throw (the headline connect-time fix)
- `AbortError` mid body-read → keeps what was collected, no throw
- non-OK status → `HttpError` carrying status + body
- a non-abort error propagates (not swallowed)
- a UTF-8 char split across two chunks decodes correctly (concat-then-decode)

**Minor — `parseLogSse` could emit a non-`LogEntry`.** A bare JSON primitive/array/null (`data: 42`) was pushed straight into `LogEntry[]`. Now kept verbatim under `raw` instead, guarded by an object check. Added tests for number/string/array/null.

**Minor — truncation left a garbage trailing frame.** A byte-cap cut usually slices the final frame mid-JSON. `collectLogSnapshot` now drops the last (partial) entry when `truncated` is set, so callers never see a corrupt half-JSON item; `truncated: true` already signals output was cut. Added tests.

Full suite: 272 pass; lint + prettier clean.

justinwlin:
Addressed the Round-2 polish (feature kept in this PR per the review discussion).

**Changes (e50409d):**
- `items.pop()` truncation: added a comment that the pop is unconditional and can drop one valid line at an exact frame boundary — acceptable since `truncated:true` already flags the loss.
- Added an end-to-end `streamLogsReply` test: v1 backend → 501 notice without touching the stream; v2 + HttpError → `{error,status}` reply (previously asserted only at the reader level).
- URL-encoded the resource ids in the `stream-pod-logs` / `stream-worker-logs` request paths (+ changeset).

The `RUNPOD_REST_VERSION`-in-`serverInfo.version` addition is intentionally kept here; the changeset documents that it changes the version string off strict semver.

FYI: `v2-rest.runpod.io` is now live (returns 200), so these v2 log-streaming tools now work against prod.

donovanclarke (approved):
lgtm. Seems to have been some scope creep? Noticing three seperate concerns being addressed in one PR. 

We can leave as-is for review ease.

PR #53, opened 2026-07-07
feat(templates+pods+catalog): persist dockerStartCmd, deploy pods from templates, GPU availability (E-3717/API-385)

justinwlin:
## Summary

Closes the template gaps in [CON-741](https://linear.app/runpod/issue/CON-741/mcp-template-dockerstartcmd-is-not-persisted-and-template-based-pod) **/** [CON-741](https://linear.app/runpod/issue/CON-741/mcp-template-dockerstartcmd-is-not-persisted-and-template-based-pod) (MCP couldn't set a template's start command, and couldn't deploy a pod from a template), plus related fixes that surfaced while verifying it.

### 1\. Template start command persists (the original bug)

`create-template`/`update-template` accepted `dockerStartCmd` but the v2 mapper dropped it, so templates came back with `args: ""`. `dockerStartCmd` is now mapped to the v2 template's `args` (a single string; a multi-element array is space-joined). `update-template` accepts it too. (v2 has no separate entrypoint field, so `dockerEntrypoint` is still not persisted on v2 — documented in the tool.)

### 2\. Deploy a pod from a template — `create-pod --templateId`

v2 `CreatePodRequest` has no `templateId`, so `create-pod` now fetches the template and spreads its container config (image, start command, ports, env, disk, volume, registry credential) into the pod body as **defaults** — any field you also pass explicitly overrides the template. `imageName` is optional when `templateId` is given. Template-based deploy is v2 + GPU only for now; clean errors are returned for v1, CPU, or when neither image nor template is supplied.

### 3\. API errors surface cleanly

`create-pod` previously only mapped a `501` to a reply and re-threw everything else, so a `400` (capacity/validation) or `5xx` bubbled out as a raw stack. Any `HttpError` now returns a readable `{error, status}` (the 501 CPU hint is preserved).

### 4\. Realtime GPU availability on the v2 catalog

`list-gpu-types` now requests `?include=AVAILABILITY` by default: each GPU carries an `availability` summary (HIGH/MEDIUM/LOW/NONE), results sort highest-stock-first, and `includeUnavailable` is a real filter (hides out-of-stock by default) instead of the previous no-op. A new `includeAvailability` param (default true) can opt out. `get-gpu-type` returns the per-datacenter availability breakdown (and URL-encodes space-containing ids) so a caller can pick a `dataCenterIds` with stock before deploying.

## Testing

* Unit + handler-integration tests for every path above (mapper round-trips, template merge + overrides, error surfacing, availability filter/sort/opt-out, per-DC get). Full suite green; type-check, lint, prettier clean. Changesets included.
* **Live (against prod v2 REST):** `create-template`/`get-template` round-trip `args`; `create-pod --templateId` fetches the real template, merges it, and posts a request the v2 API accepts (reaches scheduling) and now returns a clean reply; `list-gpu-types`/`get-gpu-type` return real availability + per-datacenter stock.

## Known limitation (backend, not this PR)

A full live pod **boot** from `templateId` could not be completed on the test account: the v2 pod scheduler returns `"no instances available"` for every GPU/datacenter — **even combos the v2 availability endpoint itself reports as in-stock** — while the GraphQL path booted a pod instantly. So the templateId request path is verified up to scheduling; the final allocation is blocked by a v2 endpoint/account issue (availability endpoint and scheduler disagree), which is worth a separate backend ticket.


justinwlin:
## Review follow-up

Addressed the nits and test gaps from review (commit 8f82360). All findings were minor/nit — the bug fix itself was confirmed correct.

**1. Stale `update-template` description.** It still read `(name, image, ports, env, readme)` and didn't mention the start command that this PR adds. Updated to list start command + registry auth, and to note omitted fields are left untouched.

**2. Lossy array→string join wasn't documented.** `["sh","-c","echo hi"]` collapses to `"sh -c echo hi"` (v2 has a single `args` string). Added the caveat to both `dockerStartCmd` descriptions, including the `sh -c` case, so an LLM reading the schema knows to pass a shell command as one pre-quoted element.

**3. Can't clear `args` on update.** Empty and absent both omit the key (correct for "don't clobber"), but there's no way to reset a command to empty. Documented this in the update-side description.

**4. Test gaps.** Added tests for the update path that motivated the design:
- empty input on update → no `args` key (the clobber-prevention that was previously only verified on create)
- empty update object → empty body
- the lossy `sh -c` collapse, so the behavior is pinned

39 mapper tests pass; type-check, lint, and prettier clean.

justinwlin:
## Added: template-based pod deploy (second half of the ticket)

The ticket (API-385 / E-3717) had **two** gaps. This PR originally fixed only #1 (`dockerStartCmd` not persisted). Commit df6625c now also closes #2 — `create-pod` couldn't deploy from a template.

**Why it needs an MCP-side bridge:** the v2 `CreatePodRequest` has no `templateId` field (a v2 pod is created from an inline `ContainerConfig`), so there's nothing to forward. Instead, when `templateId` is passed, `create-pod` fetches the template (a v2 template GET returns a full `ContainerConfig` — image, args, disk, ports, env, registry, mounts) and spreads it into the pod body as **defaults**. Any field you also pass explicitly overrides the template. `imageName` is now optional when `templateId` is given.

Because the merge is v2-`ContainerConfig`-shaped, template deploy is **v2 + GPU only** for now; the tool returns clean errors for:
- v1 (`501` — set `RUNPOD_REST_VERSION=v2`)
- CPU + templateId (`400` — CPU routes to v1)
- neither `imageName` nor `templateId` (`400`)
- a template that fails to load (surfaces the GET status/message, fires no pod POST)

**Tests**
- `podBodyFromTemplate` unit tests: picks the ContainerConfig subset + name, drops template-only fields (`id`/`serverless`/`public`/`category`), drops `null` (e.g. unset `registry`), keeps `args: ""`.
- handler integration tests: templateId → template GET then pod POST with merged body; explicit params override template; bad templateId error; v1 501; missing image/template 400; CPU+templateId 400.

**Live validation (read-only + free — no pod spun up):**
- Confirmed a real v2 template GET returns the exact `{id,name,image,args,disk,ports,env,registry,mounts,category,serverless,public}` shape the merge relies on.
- Created a template with `args: "python -u handler.py"`, re-fetched it (start command persisted), and confirmed `podBodyFromTemplate` produces a spec-valid pod body carrying the start command; deleted it (`204`). I did not create an actual pod (costs money), but the resulting POST body is a valid `CreatePodRequest` (required `name`+`image` present, `gpu` from the caller, `args`/`disk`/`ports`/`env` in ContainerConfig).

Full suite: 263 pass; type-check, lint, prettier clean. Bumped the changeset to `minor` (new capability + optional `imageName`).

justinwlin:
Addressed the Round-2 items and retitled the PR to reflect its scope (templates + pods + catalog).

**Changes (aa533cf):**
- Fixed the GPU-create `501` mislabeled as "CPU not supported" — CPU pods route to v1 before the v2 POST, so a 501 here is a GPU request; now surfaces the API's own error. Updated the two tests that locked in the mislabel.
- Documented the two low-risk template-merge assumptions inline: `podBodyFromTemplate` copies a registry credential / mounts verbatim into a body the normal create path can't produce (unvalidated against live v2, could 422), and a split `pods=v2/templates=v1` override yields a v1 template shape with no `image`.
- Changeset now calls out the `list-gpu-types` behavior change (out-of-stock hidden + reordered by stock; pass `includeUnavailable:true` for the full list).

268 tests pass.

donovanclarke:
**The registry-credential injection path is unverified and un-overridable.** podBodyFromTemplate copies the template's registry verbatim into the pod POST, but create-pod's own mapper never emits a registry field and the tool exposes no registry param? So if a template carries a registry credential, the caller cannot override or drop it, and the resulting body shape has never been accepted by the live v2 pod-create endpoint (the PR's live boot was blocked by the backend "no instances available" issue). 

This is the one path I'd want gated or validated before relying on it. 
Minimum: mention in the tool description that a template's registry credential is applied and can't be overridden here. Better: omit registry until the shape is confirmed against live v2.

**Split per-resource version override.** The template is fetched via backendFor('templates'), so RUNPOD_REST_VERSION_TEMPLATES=v1 + pods v2 yields a v1-shaped template (imageName, not image) → merged body has no image → 422. Narrow, but since template-based deploy is defined as v2-only, consider fetching the template from v2 explicitly rather than through independent resolution. _This was documented, so fine to defer if wanting to._

**list-gpu-types default behavior change.** It now hides out-of-stock GPUs and reorders by stock by default. This is a real contract change for any consumer doing a capacity/price survey. This is called out and includeUnavailable:true restores the full set, but note even then results are reordered by stock (former ordering is lost). I think we just make sure downstream consumers see it.

justinwlin:
Thanks for the review — went back through all three, and one of them changed my mind on the earlier pass. Summary of where each landed:

## 1. Template registry credential — now an operational parameter

The injection is a problem, agreed — but `registry` is a valid `CreatePodRequest` field for v2, so the issue was only that it was un-overridable, and we need it for private-image template deploys. So it's now an operational parameter: `create-pod` takes `containerRegistryAuthId`. A `templateId` deploy inherits the template's registry as a default, and the param overrides it (or sets one on a plain `create-pod`).

Opt-out too: passing an empty string emits `registry: null` (v2 accepts null), clearing a template's inherited credential — so omit = inherit, value = override, empty = none.

Verified against the live v2 API — a valid string and `null` both pass schema validation and reach the scheduler, while a wrong-typed value returns `422 $.registry: got number, want null or string`, confirming it's a real, validated field.

## 2. Split per-resource version override → fixed

`create-pod-from-template` now fetches the template from v2 explicitly (`restV2Base`) rather than the per-resource resolver. Template deploy is v2-only, so a split override (e.g. templates pinned to v1) can no longer hand back a v1-shaped template that merges into a v2 body and 422s on the missing `image`.

## 3. list-gpu-types default → reversed the hide

Agreed the default-hide made a discovery tool lossy. It now returns the full catalog by default — annotated with the `availability` summary and sorted highest-stock-first, but nothing hidden. `includeUnavailable: false` opts into deployable-only. Also dropped the `unknown` sentinel GPU on the v2 path (v1 already did), which the no-hide default would otherwise leak.

Verified live: default returns 49 GPUs, `includeUnavailable: false` returns 36, sorted HIGH→NONE.

---

Also did a pass tightening the verbose comments across these files, and documented that template fields apply whole-field (an explicitly-passed field replaces the template value rather than merging). Tests: 273 pass / 0 fail, lint + tsc clean.

donovanclarke (approved):

PR #58, opened 2026-07-27
fix(serverless): send the reshaped v2 /v2/serverless write schema

justinwlin:
`/v2/serverless` changed its create/update contract. `create-endpoint` and `update-endpoint` were sending the old shape and getting a `422`. This sends the new one.

> ⚠️ **Do not merge/release until production serves the new schema.** This targets the new shape *only* — there is no fallback. Against a host still on the old one, endpoint create/update fail (see below). Production has not promoted yet.

### What changed upstream

| | old | new |
|---|---|---|
| `type` | not on create | **required**, `QUEUE` \| `LOAD_BALANCER` |
| `scaling` | `{type, value, idleTimeout}` | `{type, queueDelay}` or `{type, requestCount}` |
| `idleTimeout` | inside `scaling` | moved to `workers` |

### Tool surface

Unchanged for existing callers — `scalerType` / `scalerValue` / `idleTimeout` keep their names and map onto the new shape. `scalerValue` defaults to `4`; passing it alone on update keeps the endpoint's current scaler.

New: **`endpointType`** on `create-endpoint` for queue-based (default) vs load-balancing routing. Fixed at creation.

`get-endpoint` / `list-endpoints` now point at `requestUrls` for an endpoint's URLs rather than having callers assemble them.

---

<details>
<summary>The 422 this fixes, and the one it introduces</summary>

Old body → new host (what `main` does today):
```
422 ["$: missing property 'type'",
     "$.scaling: missing property 'queueDelay'",
     "$.scaling: additional properties 'value', 'idleTimeout' not allowed"]
```

New body → old host (what this branch does against prod today, measured):
```
422 ["$.workers: additional properties 'idleTimeout' not allowed",
     "$.scaling: additional properties 'queueDelay' not allowed"]
```

Nothing is created by that rejection — validation runs before provisioning, and a leftover sweep after the probe came back empty.

Business-rule `422`s pass through verbatim, e.g. `"idleTimeout does not apply to queue-based endpoints scaling on requestCount"` and `"scaler field does not match endpoint type…"`. The API's own sentence is the useful one.

</details>

<details>
<summary>Test plan</summary>

312 tests; tsc, eslint, prettier, build clean.

Live end to end through the real handlers against dev (which has the new schema):

| check | |
|---|---|
| create `QUEUE` with defaults | ✅ |
| create `LOAD_BALANCER` | ✅ `requestUrls.base` + `health` |
| `get-endpoint` → `type` + `requestUrls` | ✅ |
| update `scalerValue` alone | ✅ keeps current scaler |
| switch `QUEUE_DELAY` ⇄ `REQUEST_COUNT` | ✅ both directions |
| `idleTimeout` on a `REQUEST_COUNT` queue endpoint | ✅ informative 422, verbatim |
| `LOAD_BALANCER` + `QUEUE_DELAY` | ✅ informative 422, verbatim |
| delete | ✅ 204 |

All `mcptest-*` swept; both accounts verified clean.

</details>

<details>
<summary>History: this started as a dual-shape negotiation</summary>

The first version of this PR supported both generations, negotiating per host: send the new shape, and on a shape-complaint `422` retry with the old one and remember the answer. That let it merge before promotion.

Simplified to a single shape on the call that dev's changes will reach production first. Removed with it: `src/_shared/serverless-schema.ts`, the per-host memo, the rejection classifier, `RUNPOD_SERVERLESS_SCHEMA`, the `LOAD_BALANCER`-on-old-host refusal, and 28 negotiation tests.

The tradeoff is the ⚠️ at the top: less machinery, but the merge is now gated on promotion instead of being safe at any time.

</details>

---

**Interacts with #59** (the spec resync). Two conflicting files, both intentional:

- `src/_shared/mappers.ts` — #59 carries a ⚠️ note saying these mappers do not match the vendored spec for endpoints. **This PR makes that false — take this side and drop the note.**
- `README.md` — both add a `###` section near the version-config notes. Keep both, order them.

`tests/fixtures/v2-openapi.yaml` and `tests/spec-parity.test.ts` are deliberately **not** in this PR — re-vendoring the spec is #59's job, and doing it here dragged in unrelated tag/delegation allowlist churn.

- Plugin docs counterpart: runpod/runpod-plugins-official#38
- Linear: [CON-817](https://linear.app/runpod/issue/CON-817/runpod-mcp-official-plugin-v2-v2serverless-write-schema-break-endpoint)

### DIFF
--- a/src/tools/endpoints.ts
+++ b/src/tools/endpoints.ts
@@ -12,6 +12,31 @@ import { logStreamParams, streamLogsReply } from './logs.js';
 // list/get/delete route through the adapter on both versions; create/update
 // branch on backend.version because the request model differs fundamentally.
 
+// Autoscaling bounds, mirroring the v2 spec so a bad value is a clean client-side
+// error instead of a server 422. `scalerValue` feeds two different union variants
+// with different bounds — QUEUE_DELAY takes a float >= 0.5, REQUEST_COUNT an
+// integer >= 1 — so the schema can only enforce the looser floor. The
+// REQUEST_COUNT half is checked per-call by scalerValueError() below, which needs
+// the resolved scaler type.
+const MIN_QUEUE_DELAY = 0.5;
+const MIN_REQUEST_COUNT = 1;
+const MIN_IDLE_TIMEOUT = 1;
+const MAX_IDLE_TIMEOUT = 3600;
+
+// Returns a message when scalerValue is illegal for the scaler it will be sent as,
+// or undefined when it is fine (including when either input is absent — the
+// mapper's defaults are always in range).
+function scalerValueError(
+  scalerType: 'QUEUE_DELAY' | 'REQUEST_COUNT' | undefined,
+  scalerValue: number | undefined
+): string | undefined {
+  if (scalerType !== 'REQUEST_COUNT' || scalerValue === undefined) return;
+  if (!Number.isInteger(scalerValue) || scalerValue < MIN_REQUEST_COUNT) {
+    return `REQUEST_COUNT scales on whole in-flight requests per worker, so scalerValue must be an integer >= ${MIN_REQUEST_COUNT} (got ${scalerValue}). For a fractional target use scalerType QUEUE_DELAY, which takes seconds >= ${MIN_QUEUE_DELAY}.`;
+  }
+  return;
+}
+
 export function registerEndpointTools(
   server: McpServer,
   rt: ToolRuntime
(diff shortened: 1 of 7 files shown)


src/tools/endpoints.ts
brodykellish, line 244: rewriting the 422 -> 400 seems unnecessary - but not a blocker

brodykellish (approved):

PR #59, opened 2026-07-27
feat!: resync the tool surface with the current v2 REST spec (remove tags, add ECR delegations)

justinwlin:
Brings the tool surface back in line with the live v2 spec. Four changes, all driven by re-vendoring `tests/fixtures/v2-openapi.yaml`.

### Removed — 7 tag tools

`/v2/tags` is gone from the API (`404 "The requested path was not found."` on prod and dev). The tools couldn't succeed for anyone.

⚠️ **Breaking for MCP consumers**, but the upstream removal is **confirmed intentional** — v2 REST is still in beta, so resources can be dropped. No need to wait on that question.

Still `major` in the changeset: seven tools that shipped in `@runpod/mcp-server@2.0.0` disappear, and that's breaking for anyone scripting them regardless of the API's beta status.

### Added — 3 ECR delegation tools

`/v2/registries/delegations` is new. It's a *second* way to pull a private image, not a rename:

- `create-container-registry-auth` → username + password, any registry
- `create-registry-delegation` → **AWS ECR only, stores nothing**; register a repo ARN, Runpod gets scoped pull access

### Added — network volume storage tier

`create-network-volume` gains `volumeType` (`STANDARD` | `HIGH_PERFORMANCE`). The field was in the spec all along; no tool exposed it, so HPS volumes were uncreatable here. Size range also corrected to 10–4096 GB (was documented 1–4000).

### Enabled — `list-endpoint-releases`

Implemented but commented out because the endpoint used to be dev-only. Prod now returns 200.

Plus: `create-template` stops inventing `category: NVIDIA` — v2 made it optional with that same server default.

**Surface: 52 → 49 tools.**

---

<details>
<summary>Evidence</summary>

```bash
curl -s https://v2-rest.runpod.io/v2/tags -H "Authorization: Bearer $KEY"
# 404 {"detail":"The requested path was not found."}
curl -s https://v2-rest.runpod.io/v2/templates -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer $KEY"
# 200   (control)
```

Spec operation counts: 48 → 44. `48 − 7 tags + 3 delegations = 44`, so that's the whole operation-level delta.

Tag ops were live as recently as 2026-07-06 (3 paths present in the vendored snapshot at 45, 47 and 48 ops), so this is a removal from a working state.

</details>

<details>
<summary>Test plan</summary>

301 tests pass; tsc, eslint, prettier, build clean. New wire-locks for all 3 delegation tools, `volumeType`, and releases; all v2-only tools in the uniform v1→501 sweep.

Live through the real handlers, prod and dev:

| check | result |
|---|---|
| no tag tool registered | ✅ |
| `list-registry-delegations` | ✅ `{delegations:[]}` unwrapped |
| `create-registry-delegation`, bogus ARN | ✅ `400 "Invalid ARN format"` |
| `delete-registry-delegation`, unknown id | ✅ `400 "Delegation not found or access denied"` |
| `list-endpoint-releases` on prod | ✅ 200, `endpointVersion=12` |

Those 400s are the point: a real handler validating input, not `404 path not found`. Nothing created.

</details>

<details>
<summary>Reviewed and deliberately unchanged</summary>

- **Newly strict schemas.** The refresh added `additionalProperties: false` widely. The two identity-mapper tools (`create-container-registry-auth`, `update-network-volume`) match their schemas exactly.
- **`GpuManufacturer`** is response-only (`GpuType.manufacturer`).
- **No new query parameters** on any existing operation.
- **The serverless endpoint contract** also changed in this spec — that's #58's job, not this PR's. See the note below.

An adversarial review of this PR caught the `volumeType` gap: I'd claimed the field was absent from the spec, having concluded that from a diff, where an *unchanged* field correctly shows nothing. Fixed above.

</details>

---

**Interacts with #58** (the serverless write shape). Two conflicting files, both intentional:

- `src/_shared/mappers.ts` — the ⚠️ note in this PR says the endpoint mappers don't match the vendored spec. #58 fixes exactly that, so **when #58 lands, take its side and drop the note.**
- `README.md` — both add a `###` section near the version-config notes. Keep both, order them.

#58 no longer re-vendors the spec or touches the parity gate, so this PR owns both outright.

- Plugin docs counterpart: runpod/runpod-plugins-official#39 — merge this first
- Linear: [CON-817](https://linear.app/runpod/issue/CON-817/runpod-mcp-official-plugin-v2-v2serverless-write-schema-break-endpoint)


brodykellish (approved):

lukepiette (commented):
Reviewed with independent verification — this holds up well.

**Verified independently:**
- **The tags 404 is real and current.** Coincidentally hit this in the field today before reading this PR: `create-tag` through the hosted server, with a *freshly minted* valid token, returned `404 "The requested path was not found"`. So the seven tag tools were already dead for every caller; removal + `major` changeset is the right call.
- **Ran the branch locally:** 293 pass / 0 fail; lint/type-check clean.
- **Spec claims check out** in the vendored yaml: `/v2/registries/delegations` and `/{id}` are present, and the network-volume size bounds are `minimum: 10` / `maximum: 4096` as documented.
- Delegation implementation details are careful where it counts: `name` omitted (not nulled) under `additionalProperties: false`, `encodeURIComponent` on the delegation id, the `{delegations:[…]}` envelope read directly instead of forcing it through `backend.unwrap`, and uniform v1→501 notices. The README section contrasting credentials vs delegation is genuinely useful.

**Minor comments (none blocking):**
1. `create-network-volume`'s `size` is still `z.number()` with the range only in prose. Now that the bounds are settled (10–4096), consider `.int().min(10).max(4096)` — client-side validation saves agents a failed API roundtrip, and the schema is the contract agents actually read. Same consideration for `update-network-volume`.
2. README: the v2 migration-note blockquote got indented under the bullet list (`- …` → `  > ⚠️ …`). Assuming that nesting is intentional — worth a quick render check.
3. **Merge coordination:** this conflicts (trivially — tool-registration snapshot, spec-parity allowlist, `tools.ts` registrar list, handlers tests) with in-flight #60 and #62, which add tools to the same frozen lists. Any land order works; happy to rebase mine after this goes in, since #59 owns the spec fixture.

One process note worth keeping from this PR: the "adversarial review caught my own volumeType claim" disclosure in the description is exactly the right habit — diff-reading hides unchanged fields, and saying so saves the next reviewer from the same trap.

PR #60, opened 2026-07-28
feat: add Hub discovery/deploy, Public Endpoints, and GPU SKU pinning tools

lukepiette:
## Summary
Four new tools bringing Hub discovery/deploys, Public Endpoint discovery, and GPU SKU pinning to the MCP. The catalogs are served by the public GraphQL endpoint (no auth); deploys and GPU updates use the authenticated GraphQL `saveEndpoint` mutation via a new `graphqlAuthed` runtime helper (API-key Bearer + variables). All are version-agnostic — same path on v1 and v2, since none of these have a REST home yet.

### `list-hub-repos`
- Lists repos published to the Runpod Hub (prebuilt Serverless workers and Pod templates, e.g. vLLM, ComfyUI, Axolotl) via the `listings` query.
- Each result includes repo metadata (stars, deploys, category, tags, hub/GitHub URLs) plus the currently listed release with its `hubReleaseId` and prebuilt `imageName`. Sorted most-deployed first.
- Filters: `searchTerm`, `category`, `type` (SERVERLESS/POD), `repoOwner` — client-side, since the public GraphQL path does not accept variables. Opt-in `includeConfig` adds the release's parsed `.runpod/hub.json` config (excluded by default; it can be tens of KB per listing).

### `deploy-hub-repo`
- Deploys a Hub repo's listed release as a new Serverless endpoint — the same operation as clicking Deploy on the Hub. Identify the repo by `repo` ("owner/name") or `hubReleaseId`.
- The release supplies the prebuilt image, container disk, CUDA constraints (`minCudaVersion` computed as the lowest allowed), and env-var defaults (booleans serialized through `trueValue`/`falseValue`); caller overrides for `env`, `gpuIds`, `gpuCount`, `containerDiskInGb`, workers, scaling, and `flashboot`.
- Fails fast **before anything is created** when required env vars are missing, the listing is a POD, the repo is unknown, or no GPU pool is specified/provided.
- `gpuIds` documents the SKU-exclusion syntax (`'AMPERE_16,-NVIDIA RTX A4500'`) so agents can pin SKUs at deploy time.

### `set-endpoint-gpus`
- Sets which GPUs **any** Serverless endpoint's workers run on — including pinning specific GPU SKUs, which the REST API cannot express (pool-level `gpuPoolIds` has no exclusion concept).
- Pass a raw `gpuIds` string, or `pools` + `excludeGpuTypeIds` (GPU type ids from `list-gpu-types`) and the `POOL,-SKU` string is built automatically; excluding all but one SKU in a pool pins that SKU exactly.
- `saveEndpoint` is not a sparse update — omitted endpoint-level scalars reset to server defaults (verified live: an id+name+gpuIds-only update reset workersMax 1→3 and idleTimeout 5→10). The tool reads the endpoint's current settings first and echoes every field back with only the GPU selection changed, preserving workers, scaling, timeouts, FlashBoot, locations, and network volumes.

### `list-public-endpoints`
- Lists Runpod Public Endpoints (managed, pay-per-use model APIs) via `allAiApiPublicConfigs`.
- Each result includes the `endpointId` for `run-endpoint`/`runsync-endpoint` (or `https://api.runpod.ai/v2/{endpointId}`), plus model name, modality, owner, and pricing parsed from the catalog `metadata` JSON (defensive parse). Live-only by default (`includeOffline:true` opts in), with `searchTerm`/`modality`/`owner` filters.

All list tools use the shared pagination envelope; all four tools are registered in the frozen tool-surface snapshot and allowlisted in the v2 spec-parity gate. Changesets included (minor).

## Test plan
- [x] `pnpm test` — 316 pass / 0 fail, including 21 new offline handler tests pinning the outbound GraphQL queries and mutation inputs (hubReleaseId, config-derived hardware, env default/override/boolean/passthrough merge, the read-then-echo field preservation on set-endpoint-gpus, pools+exclusions string building), Bearer auth present on authenticated calls and absent on public queries, all client-side filters, the fail-fast guard rails (no mutation sent), and the pagination envelopes.
- [x] `pnpm lint`, `pnpm type-check` clean; touched files pass `prettier --check` (7 files fail prettier on master pre-existing, untouched).
- [x] Live end-to-end against the built `dist/stdio.mjs` over MCP stdio:
  - **Hub list**: full 71-listing catalog; `searchTerm: "axolotl"` returns the `hubReleaseId` and `registry.runpod.net` image matching the console's deploy payload; `includeConfig`, filters, and cursor pagination verified.
  - **Hub deploy**: vLLM without `MODEL_NAME` fails fast (nothing created); ComfyUI deployed for real, verified via `get-endpoint`, then deleted. Subsequently used to stand up a 24-endpoint ComfyUI fleet (one per GPU SKU) with the exclusion syntax.
  - **GPU pinning**: confirmed live that a minimal `saveEndpoint` update resets omitted scalars (workersMax 1→3, idleTimeout 5→10) — the read-then-echo design exists because of this. `set-endpoint-gpus` then pinned an endpoint to a single SKU via `pools` + `excludeGpuTypeIds` with workers/scaling/timeouts/template verified untouched.
  - **Public endpoints**: 41 live (50 with `includeOffline`); search/modality/owner filters verified.


justinwlin:
Pushed 4 commits to this branch rather than leaving a list of asks (plus one prettier commit). Suite goes 324/316 → **337/329, 0 failures**; build, type-check and lint clean.

One blocker, three should-fixes, and a test gap that hid the blocker. Each with the evidence, since a couple of these are only visible against the live schema.

---

## 1. `set-endpoint-gpus` failed on every endpoint with a network volume

**Gap.** The read selected `networkVolumeIds { networkVolumeId dataCenterId }` and echoed those objects into `EndpointInput`. Read and write are different types whose names differ by one suffix:

- read → `NetworkVolumeIds` — `networkVolumeId` + `dataCenterId`
- write → `NetworkVolumeIdsInput` — `networkVolumeId` **only**

The server rejects the read shape outright, so the tool failed 100% of the time on any endpoint with a volume attached. Endpoints without one echo `null` and work fine, which is why live testing came back clean.

**Fix.** Map to the input shape explicitly.

**Why the guard.** A test now asserts the entry key set is exactly `['networkVolumeId']`, so the read shape can't get passed through again.

<details>
<summary>Probe evidence</summary>

GraphQL coerces variables before resolvers run, so an unauthenticated request reveals whether a field exists — `Unauthorized` means it validated, `BAD_USER_INPUT` means it didn't. Three requests:

```
networkVolumeId + dataCenterId  →  Field "dataCenterId" is not defined by type
                                   "NetworkVolumeIdsInput"        [BAD_USER_INPUT]
networkVolumeId alone           →  Unauthorized      (passed coercion)
networkVolumeId + made-up field →  Field "totallyFakeField" is not defined by type
                                   "NetworkVolumeIdsInput"        [BAD_USER_INPUT]
```

`dataCenterId` is rejected identically to a field I invented. On the read side it's legal — a made-up field there fails with `Cannot query field ... on type "NetworkVolumeIds"`.
</details>

## 2. Five echoable fields were missing from the echo

**Gap.** `templateId`, `allowedCudaVersions`, `minCudaVersion`, `compliance` and `modelReferences` are all readable on `Endpoint` and writable on `EndpointInput`, but weren't in the read selection or the input.

Worth saying: **they don't currently reset.** I measured the actual reset set and it's narrower than "every omitted field" — an `id+name+gpuIds` update resets `workersMax`, `idleTimeout` and `scalerValue`, and nothing else. Your note that the template survives is correct.

**Fix.** Echoed anyway, but only when non-null — omitting a field that already reads `null` resets it to a default that's also `null`, while sending an explicit `null` risks a type rejection for no gain. The header comment now records the measured reset set rather than a general claim.

**Why bother, if they don't reset.** Cheap insurance on undocumented server behaviour: the reset set isn't specified anywhere, it can widen without warning, and this tool's entire design assumes the echo is complete. Two shapes matter if anyone extends this later — `compliance` is a `[Compliance]` **enum** (rejects `'gdpr'`, wants `'GDPR'`), so read values must pass back verbatim, and `allowedCudaVersions` is a plain comma-separated `String` on **both** sides, not a list.

<details>
<summary>Live measurement on a throwaway endpoint</summary>

Created a scratch endpoint with deliberately non-default settings, ran the minimal update, then deleted it. No existing endpoints were touched.

```
minimal update (id + name + gpuIds):
  workersMax   7  -> 3
  idleTimeout  42 -> 10
  scalerValue  9  -> 4
  templateId / allowedCudaVersions / minCudaVersion / compliance /
  modelReferences / executionTimeoutMs / scalerType / flashBootType /
  gpuCount / workersMin / locations   ->  unchanged

full echo (the fixed payload), gpuIds ADA_24 -> ADA_48_PRO:
  gpuIds changed. nothing else changed.
```
</details>

## 3. The API key was following the public-discovery override

**Gap.** `graphqlAuthed` attaches the caller's key as a Bearer token but resolved its host from `publicGraphqlBase` — i.e. `RUNPOD_PUBLIC_GRAPHQL_URL`.

That variable is documented as the no-auth discovery override for `list-gpu-types`/`list-data-centers`, and the repo points it at stubs freely (`tests/backend.test.ts` uses `localhost:8`, `scripts/local-http.ts` targets a dev backend). So "safe to point anywhere" quietly became "send the caller's credential there" — and on the hosted server that credential is a per-user OAuth-minted key, not a shared one.

**Fix.** New `authedGraphqlBase` / `RUNPOD_AUTHED_GRAPHQL_URL`, used only by `graphqlAuthed`. `RUNPOD_PUBLIC_GRAPHQL_URL` keeps its no-credentials guarantee. A test asserts the authed calls never follow the public var.

**Why a new var rather than reusing `RUNPOD_GRAPHQL_URL`.** #61 is separately turning that one into its credential-check target; sharing it would leave the variable meaning two different things depending on merge order.

Also updated the CLAUDE.md architecture paragraph, which still described the GraphQL API as public, no-auth and read-only — no longer true with a billable `saveEndpoint` mutation on it.

## 4. `buildHubEnv` — three ways the env merge misfired

**An empty override satisfied a required key.** The guard tested `key in overrides`, so `env: { MODEL_NAME: '' }` submitted an empty string and reported nothing missing. The fail-fast check passed and the deploy produced a broken worker — the exact outcome the guard exists to prevent. Empty is now treated as not supplied.

**Caller booleans skipped the `trueValue`/`falseValue` mapping** that defaults got. An agent filling in a boolean input naturally sends `'true'`, and the worker received `'true'` where its own schema declares `'1'`. Now applied to overrides too, for exact `'true'`/`'false'` only, so a deliberate literal still passes through.

**An optional key with no default was submitted as `''`.** That shadows the image's own fallback — `os.environ.get('X', 'fallback')` returns `''` when `X` is set to `''`. Now omitted entirely.

## 5. `list-hub-repos` paging could skip and duplicate

**Gap.** Sorted on `deploys` alone. Each page refetches the whole catalog, `listings(input: {})` promises no ordering, and sort stability only preserves whatever order the server happened to return — so the long tail of 0-deploy repos moves between requests, and page 2 re-shows or skips entries.

**Fix.** Tiebreak on `id`. A test feeds the same tied listings back in a different order and asserts the page comes out identical.

## 6. The tests couldn't see any of the above

**Gap.** The `set-endpoint-gpus` `assert.deepEqual(input, {...})` was a change-detector — the expected object was a copy of the implementation's own literal. It passed with fields missing from *both* sides at once, and passed with the `networkVolumeIds` shape the live server rejects, because the fake fetch echoes anything back. Its comment claimed to pin that every value is carried over verbatim, which was the one thing it didn't check. `buildHubEnv` and `minCudaVersion` were exported with no direct tests.

**Fix.**
- Echo assertions derived **from the fixture**: every fixture key except `gpuIds` must return unchanged, the input key set must match exactly, `networkVolumeIds` entries must have exactly one key. Add a field to the fixture and the test demands it be echoed.
- Second fixture endpoint covering the all-nulls case, pinning the omit-when-null path.
- `tests/hub-env.test.ts` — 11 direct unit tests for the two pure helpers.
- The list-`*` pagination gate iterated a hardcoded 5-name array while claiming to cover every list tool, so both new list tools were unguarded. Now derived from the registered surface.
- `ALLOWLIST_UNMAPPED_TOOLS` split into `ALLOWLIST_NEVER_REST` and `ALLOWLIST_PENDING_REST`. The eight runtime entries are permanent; the four new ones are "not yet". Merged they read identically, and a "not yet" that nothing revisits becomes a permanent hole in this gate — the tool stays unguarded long after REST ships.

**Why I trust these.** I reverted each fix one at a time; all seven mutants fail at least one test.

---

All touched files pass `prettier --check` — that included one pre-existing long line at `tests/handlers.test.ts:1085`, wrapped in its own commit so it's easy to skip while reviewing.

justinwlin (approved):
Approving. The four tools are well-built — the read-then-echo design in `set-endpoint-gpus` is exactly right given `saveEndpoint`'s non-sparse behaviour, and failing fast before creating anything in `deploy-hub-repo` is the correct instinct.

I pushed fixes for the issues in my comment above rather than sending them back as asks; the substantive one was the `NetworkVolumeIdsInput` shape, which would have broken every volume-bearing endpoint. Happy for you to reword or drop any of it — in particular `RUNPOD_AUTHED_GRAPHQL_URL` is a name I picked, so rename it freely.

PR #61, opened 2026-07-28
fix(http): surface dead credentials as HTTP 401 + WWW-Authenticate so clients re-auth

lukepiette:
## Summary
Fixes the hosted server's dead-credential behavior discovered during a live session: when the OAuth-minted API key expires or is revoked mid-session, every upstream Runpod API call fails with a 401 — but the MCP SDK wraps thrown tool errors into a **200** JSON-RPC tool result, so the client never sees an HTTP 401, never re-runs its OAuth flow, and the user is stuck with bare `Unauthorized` tool errors until they notice and manually reconnect.

- The hosted request handler now **pre-flight verifies the bearer** (one authenticated `myself { id }` GraphQL query against the auth backend) before handling the MCP request. A dead credential gets a proper **HTTP 401 + `WWW-Authenticate: Bearer realm="mcp", resource_metadata="…/.well-known/oauth-protected-resource"`** — the signal OAuth-capable MCP clients (Claude Code, Cursor, etc.) use to re-authenticate automatically.
- Verdict logic pins the backend's observed live behavior: invalid/expired key → HTTP 401; anonymous token → 200 with `myself: null` (no identity → every downstream call would 401); valid key → `myself.id`.
- **Cached by token hash** (never the raw token): valid verdicts ~5 min, invalid ~30 s (so a just-re-authorized user isn't locked out). A warm instance adds no per-request latency.
- **Fails open** on auth-backend 5xx/network errors and unrecognized response shapes — a blip in the auth backend cannot take down the whole server; tools then surface the real upstream error. Escape hatch: `MCP_SKIP_CREDENTIAL_CHECK=true` (documented in CLAUDE.md / docs/context.md).
- Defense-in-depth: upstream 401s that still reach a tool (stdio env-key users, cache windows) now carry an actionable hint on the `HttpError` message ("credential expired or revoked — re-authenticate or update RUNPOD_API_KEY") instead of a bare `401 - Unauthorized`, so agents can tell "re-auth" apart from "forbidden".

Changeset included (patch).

## Test plan
- [x] `pnpm test` — 308 pass / 0 fail, including 13 new tests: verdict logic for all four observed backend responses (valid / 401 / anonymous `myself:null` / unrecognized shape), fail-open on 5xx and network errors, per-token caching with valid- and invalid-TTL expiry, and `handleMcpRequest` wiring (dead credential → 401 with the exact `WWW-Authenticate` resource-metadata header; missing bearer unchanged and never consults the checker; `MCP_SKIP_CREDENTIAL_CHECK=true` bypasses the pre-flight).
- [x] `pnpm lint`, `pnpm type-check` clean; touched files prettier-clean (pre-existing violations in `src/_shared/http.ts` untouched).
- [x] **Live end-to-end** against the built `dist/http.mjs` serving real HTTP with the real Runpod API:
  - dead key → `HTTP/1.1 401 Unauthorized` with `WWW-Authenticate: Bearer realm="mcp", resource_metadata="…/.well-known/oauth-protected-resource"`.
  - real key → `initialize` returns 200 (`text/event-stream`) and a `tools/call` (`list-data-centers`) succeeds through the pre-flight.
  - repeated dead-key request → 401 served from the verdict cache in ~20 ms (no upstream roundtrip).

## Context
Found while dogfooding the hosted server in Claude Code: the session credential died mid-session, all tools started returning bare `Unauthorized` (initially misread as a key-scope problem), and recovery required a manual `/mcp` reconnect. With this change the client re-auths automatically.

### DIFF
--- a/.changeset/credential-preflight.md
+++ b/.changeset/credential-preflight.md
@@ -1,0 +1,19 @@
+---
+'@runpod/mcp-server': minor
+---
+
+Surface dead credentials as a proper HTTP 401 on the HTTP server so OAuth-capable MCP clients re-authenticate automatically.
+
+Previously, when a bearer token was revoked mid-session (e.g. an OAuth-minted key), every upstream Runpod call failed with a 401 that the MCP SDK wrapped into a 200 JSON-RPC tool error — the client never saw an HTTP 401, never re-ran its auth flow, and the user was stuck with bare "Unauthorized" tool errors until they manually reconnected. The request handler now pre-flight verifies the bearer (one `myself` GraphQL query, cached ~60s valid / ~30s invalid by token hash, never the raw token) and answers `401` + `WWW-Authenticate` with the protected-resource metadata when the credential is dead. `WWW-Authenticate` is exposed via CORS so browser clients can read it, and a rejected credential carries `error="invalid_token"` to distinguish it from a request that sent none (RFC 6750 §3.1). `ToolContext` gains an optional `onUnauthorized` callback, invoked when an outbound call returns 401, which drops the cached verdict so the next request re-checks.
+
+The check is deliberately conservative so it can never reject a working key:
+
+- It fails **open** on anything indeterminate — auth-backend errors, 5xx, 403 (this host sits behind a WAF, and a block is not a revocation), a slow backend (time-bound), and a GraphQL response carrying an `errors` array (a resolver blip returns `myself: null` for everyone at once; treating that as invalid would 401 every valid key). Failing closed on any of these would make OAuth clients re-run the flow, minting a new API key per attempt — an unrecoverable loop.
+- It runs only for the requests the transport would actually run a tool for: a `POST` whose `Content-Type` the SDK treats as JSON and whose body invokes a method (including inside a batch). Everything else is passed through unchecked.
+- It self-disables (logging `skip_env_mismatch`) when the REST/Serverless hosts and the auth-GraphQL host disagree about environment in either direction, since it would otherwise validate a key against the wrong backend.
+- Request bodies are capped at 4 MB (matching the SDK) on hosts that do not pre-parse them.
+- The verdict cache is safe under load: eviction never disowns an in-flight check, and TTLs use a monotonic clock, so a burst of distinct tokens or a wall-clock adjustment cannot let a since-revoked key linger as `valid`.
+
+Set `MCP_SKIP_CREDENTIAL_CHECK=true` to disable the pre-flight entirely.
+
+Notes for consumers of the published `./http` export (not only the hosted deployment): the pre-flight adds one outbound `myself` call to the Runpod GraphQL host per checked request, and a new dependency on that host being reachable (it fails open if not). This does **not** close the pre-existing key-validity oracle — a caller who speaks MCP can still learn 401-vs-not for a token, one upstream call each, with no rate limiting; closing that needs rate limiting, which this does not add.
--- a/.changeset/http-401-reauth.md
+++ b/.changeset/http-401-reauth.md
@@ -1,5 +1,0 @@
----
-'@runpod/mcp-server': patch
----
-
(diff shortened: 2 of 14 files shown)


justinwlin:
Went through this closely (pushed onto `fix-http-401-reauth`, on top of your `b969ee1`). The mechanism is right and the docs are genuinely thorough — thank you for those. I ran it through several adversarial passes and fixed a set of issues locally; everything below is fixed on the branch now, each with a regression test that fails when the fix is reverted (mutation-checked).

The important ones — a few would reject a **valid** key and, because each Runpod re-auth mints a new API key, spin an unrecoverable loop:

| Area | Problem | Fix |
|------|---------|-----|
| Verdict | `{data:{myself:null},errors:[…]}` (HTTP 200 resolver blip) was classified `invalid` → 401s every valid key at once during a backend hiccup | any `errors[]` → fail open; only a clean `myself:null` is anonymous |
| Gate bypass | a mixed JSON-RPC batch (`[response, tools/call]`) skipped the gate via `every()` while the SDK still ran the tool | `every` → `some` |
| Gate bypass | `application/json-patch+json` / `application/jsonx` skipped the gate (exact media-type match, but the SDK uses `includes`) | mirror the SDK's `includes` check |
| DoS | `readJsonBody` had no size cap and ran before the method/content-type checks — unbounded buffering + a slow non-JSON body held the request slot open on the `./http` path | 4 MB cap + read only after the cheap skip checks |
| Env guard | one-directional — moving only `RUNPOD_GRAPHQL_URL` (its documented dev override) validated a prod key against a dev backend and 401'd every good credential | bidirectional: skip when auth-GraphQL and REST/Serverless disagree about environment either way |
| Cache race | at capacity, eviction popped the front unconditionally and could disown a still-in-flight check — under a token flood a since-revoked key could cache as `valid` for a full TTL | eviction skips entries with a live promise |
| Cache clock | TTLs used `Date.now`, so a wall-clock step-back could stretch the ~60s bound to minutes | monotonic `performance.now` |

<details>
<summary>Verification</summary>

- Tests 316/308 → **415/407** (0 fail, 8 opt-in skips). type-check + lint + prettier clean.
- Each fix above was mutation-tested: reintroducing the bug drops exactly one test.
- Bypasses and the DoS/slowloris were reproduced over a real socket before and after the fix.
- I squashed the two changesets into one (single never-released feature) and had it disclose the residual key-validity oracle and that the pre-flight also applies to the published `./http` export, not just the hosted deployment. Softened one README line: the *first* call after a mid-session revocation can still be a wrapped 200; the one after re-checks and 401s.
</details>

There's also a small unrelated commit on top that unbreaks the Release workflow (it's been failing since Jul 13 on `npm i -g npm@latest`; details in that commit message).

justinwlin:
One follow-up I did **not** change, because it's your call on intent:

The pre-flight sends the caller's bearer to `RUNPOD_GRAPHQL_URL`. Per `backend.ts`, key-bearing GraphQL calls are supposed to go through `authedGraphqlBase` (`RUNPOD_AUTHED_GRAPHQL_URL`) — the variable kept deliberately separate so a freely-repointable host can't capture the caller's key. `RUNPOD_GRAPHQL_URL` is otherwise the OAuth flash-auth var (no caller key) and is documented as repointable to a dev backend.

- **Today: no impact** — all three GraphQL vars default to `api.runpod.io/graphql`.
- **If `RUNPOD_GRAPHQL_URL` is ever repointed**, the pre-flight would send every caller's key to that host. The bidirectional env-guard covers the "only GraphQL moved" case (it skips), but a consistent dev override (REST + GraphQL both moved) would still send keys to the dev host.

Options: (a) point the pre-flight at `authedGraphqlBase` and compare that host in the env-guard — matches the existing convention and closes the path; or (b) leave it on `RUNPOD_GRAPHQL_URL` deliberately since it targets the same flash-auth backend the OAuth flow uses. I lean (a), but it's your feature — either is fine, just wanted it on the record.

justinwlin (approved):
Approving — reviewed in depth and the fixes above are all on the branch, tested and mutation-checked. The two items left for you: the `RUNPOD_GRAPHQL_URL` vs `authedGraphqlBase` question above, and a quick `vercel env ls production` to confirm the gate isn't self-disabling in prod (it's safe either way — it just logs `skip_env_mismatch`).

lukepiette:
Outstanding review — thank you. The resolver-blip fix and the two gate bypasses in particular would have been real incidents; the mutation-testing discipline made all of it easy to trust. All seven fixes verified locally (415/407, matching your numbers).

Both open items are now addressed on the branch (`dfbeaeb`):

**1. `RUNPOD_GRAPHQL_URL` → `authedGraphqlBase`: done, went with your (a).** The deciding argument is exactly the one you laid out: the pre-flight carries the *caller's* key, and `RUNPOD_AUTHED_GRAPHQL_URL` is the variable `backend.ts` reserves for key-bearing GraphQL precisely so a freely-repointable host can never capture caller credentials. The env-mismatch guard compares the same resolver, so the checker and guard can't disagree about defaults. All 21 test references updated.

**2. CI was red on the branch — fixed.** The `hang: true` timeout tests pended on a promise whose only wake-up was `AbortSignal.timeout()`'s **unref'd** internal timer; with nothing else keeping the event loop alive, the loop drained before the abort fired and node:test cancelled the whole file ("Promise resolution is still pending but the event loop has already resolved") on every Node version. The mock now holds a ref'd keep-alive timer until the abort lands. (Production unaffected — real fetch refs its sockets.)

Still open for a human with Vercel access: `vercel env ls production` to confirm the env guard isn't self-disabling in prod (my CLI isn't linked to the project). @lukepiette — that one's a 30-second dashboard check: if any `RUNPOD_*_API_URL` var is set to a non-default host without a matching `RUNPOD_AUTHED_GRAPHQL_URL`, the gate is currently logging `skip_env_mismatch` instead of running.

PR #62, opened 2026-07-28
feat(jobs): diagnose stuck IN_QUEUE jobs — crash-looping workers vs capacity

lukepiette:
## Summary
Closes the diagnostic gap that misled a real agent session today: a job stuck `IN_QUEUE` has two very different causes that the job status alone cannot distinguish — **no host has the endpoint's GPU available** (capacity), or **a worker spun up and is crash-looping** (the platform marks it UNHEALTHY while the job stays queued forever). In a Hub GPU test run across 24 SKUs, 4 of 5 "capacity timeouts" were actually crash-looping workers; the agent (me) confidently reported them as capacity because nothing in `get-job-status` hinted otherwise.

Two changes:

### 1. `get-job-status` auto-attaches a worker-health diagnosis (the composed fix)
When the returned status is `IN_QUEUE`, the tool best-effort fetches the endpoint's worker list and attaches:
- `workerHealth`: the summary (`{idle, initializing, running, throttled, total, unhealthy}`).
- `hint`: plain-language classification — UNHEALTHY workers → *"container is likely crash-looping … this is a worker failure, NOT a capacity shortage. Inspect with stream-worker-logs (workerId: …)"*; zero workers → capacity guidance (check `list-gpu-types`, widen gpuIds/CUDA); throttled/initializing → wait guidance.

v2-only (the workers listing has no v1 REST home; v1 returns the status unchanged), and the extra lookup is wrapped so it can **never fail the status call** — an unrecognized response degrades to the plain status.

### 2. Cross-linked tool descriptions (the breadcrumbs)
`get-job-status`, `endpoint-health`, `list-endpoint-workers`, and `stream-worker-logs` now reference each other and name the crash-loop signature (repeated `start container` system lines with no container output), so an agent that has never seen this platform behavior gets routed from "job is stuck" to the worker logs.

Changeset included (minor).

## Test plan
- [x] `pnpm test` — 301 pass / 0 fail, including 5 new offline tests: UNHEALTHY worker → `workerHealth` + crash-loop hint naming the worker id (and the exact second outbound URL); zero workers → capacity hint; terminal statuses skip the diagnosis (single outbound call); v1 → status unchanged with no second call; unrecognized workers response degrades gracefully (no crash, no hint). The pre-existing `get-job-status` golden is unaffected.
- [x] `pnpm lint`, `pnpm type-check`, prettier on touched files clean.
- [x] **Live end-to-end** against the real API using the exact endpoint that caused today's misdiagnosis (a ComfyUI endpoint with a genuinely crash-looped H200 worker): submitted a probe job → `get-job-status` returned `IN_QUEUE` + `workerHealth: {total:1, unhealthy:1}` + a hint naming worker `1tcb8lhb77hauh` and stating it is *not* a capacity shortage. Probe job cancelled afterwards.

## Context
Companion to the API-team asks already filed: a queue reason on the `/status` response and exit codes in the worker-logs stream would make this diagnosis native to the platform; this PR gives agents the same signal today from the MCP layer.


justinwlin:
Reviewed this closely — verified the workers URL, response shape (`summary` keys, `workers[].id/.status`, the `UNHEALTHY` literal) all match the v2 spec and `list-endpoint-workers`, and it's cleanly best-effort (never fails the status call). Tests pin both outbound URLs, the classification, and graceful degradation. Nice fix. One thing worth a thought before merge:

**Polling amplification.** Agents poll `get-job-status` in a loop while a job is queued. With this change, every poll during `IN_QUEUE` now fires a second call (the worker-list fetch) — so a job queued for a few minutes recomputes the same hint dozens of times, even though nothing changed between polls. It's bounded and best-effort, so not a correctness issue, but the diagnosis is really only useful once.

Options if you want to trim it (none blocking):
- only attach the diagnosis after the job has been queued a little while (skip the first N seconds / first poll), or
- briefly cache the worker summary per endpoint so rapid polls reuse it.

Everything else is minor: the zero-workers hint can read as "capacity" during a normal scale-from-zero cold start (self-corrects on the next poll — could add "or scaling up from zero"), and the `throttled`/`initializing` hint branches aren't covered by a test. Approving — these are all optional polish.

justinwlin (approved):
Approved — verified the worker URL/response shape against the v2 spec and `list-endpoint-workers`, ran the suite (0 fail), and the change is cleanly best-effort. Only non-blocking note is the polling amplification above; the rest is optional polish.

lukepiette:
All three notes addressed in `48a6230`:

1. **Polling amplification — fixed with a bounded 15s per-endpoint cache** (module-level so it survives per-request tool registries on a warm hosted instance, capped at 500 entries). Rapid polls reuse the diagnosis; a new test pins it (3 polls → exactly 1 workers fetch, hint present on all 3). Went with the cache over skip-first-N because agents poll at unknown intervals — the cache bounds the amplification regardless of cadence.
2. **Zero-workers hint** now leads with the benign case: *"may simply be scaling up from zero (normal for the first seconds of a cold start)"* before suggesting capacity/CUDA causes.
3. **`throttled` and `initializing` branches now have tests.**

343 tests / 0 fail, lint + type-check + prettier clean. Thanks for the careful pass — the amplification point was real; our own test suite polls at 10s intervals and would have fired ~70 redundant worker fetches per queued job.

PR #64, opened 2026-07-29
feat(http): make 429s actionable — name the exhausted quota window and reset time

lukepiette:
## Summary
Makes 429s actionable for agents. Today a rate-limited call surfaces as `Runpod API Error: 429 - {"detail":"rate limit exceeded for the hour window"}` — and an agent's natural response to that is retrying immediately, which digs the hole deeper. Observed live: a bulk endpoint GC (24 deletes) burned the v2 hourly quota (3,600/hr) and every delete bounced with no guidance; the calling tooling misread the failure entirely.

The v2 API already sends IETF draft `RateLimit` headers on every response (`"minute";r=178;t=44, "hour";r=0;t=1724, …`). On any 429 the HTTP client now parses that header and appends a hint naming **which quota window is exhausted and when it resets**:

```
Runpod API Error: 429 - {"detail":"rate limit exceeded for the hour window"}
  (rate limited — the hour quota is exhausted; it resets in ~1724s. Wait before retrying and pace bulk operations)
```

Falls back to generic back-off guidance when the header is absent (the v1 API sends no RateLimit headers). `HttpError.status`/`.body` are unchanged for programmatic callers. Changeset included (patch).

## Test plan
- [x] `pnpm test` — 331 pass / 0 fail; 4 new tests: hint parsing names the exhausted window + reset seconds; generic fallback with no header or no exhausted window; a 429 response carries the hint on the message while `.body` stays clean; non-429 errors carry no hint (golden unchanged).
- [x] `pnpm lint`, `pnpm type-check`, prettier clean.
- [x] Header format verified against the live v2 API (`v2-rest.runpod.io`) — `RateLimit`/`RateLimit-Policy` captured from real responses during today's rate-limit incident.

## Context
Third error-message-as-agent-UX fix from today's field session, alongside #61 (401 → re-auth signal) and #62 (IN_QUEUE → worker-health diagnosis). Related: the API team is doubling the rate limit; this hint makes whatever the limit is self-explanatory when hit.


justinwlin:
Pushed a revision that reworks the hint before review — the original inline parser could make the hint actively misleading in a common case, so it moved into its own module (`src/_shared/rate-limit.ts`) with each failure mode pinned by a test.

What changed:

- **Reports the longest wait, not the first exhausted window.** 3000/hr is an exact multiple of 60/min, so the request that zeroes the hour quota is also the 60th of its minute — both windows read `r=0` together, and the old `find(r === 0)` answered a 30-minute lockout with "resets in ~12s". That's the exact retry-now failure this PR exists to prevent.
- **Reads `Retry-After` too.** The server's explicit delay was ignored; the hint now takes the longest wait either header justifies, so it never asks for less than `Retry-After` does.
- **Tolerates legal header variation.** The old regex was pinned to `"name";r=N;t=N` in that exact order; `RateLimit` is a structured field, so reordered or extra params (`"hour";t=1724;r=0`) silently fell back to the generic hint.
- **Never invents a number.** A partially readable header hedges ("wait at least …") or says so ("no usable reset time") instead of quoting a wait the header doesn't support; implausible resets (> 7 days) and unvalidated window names are rejected rather than printed into agent-facing text.
- **Covers the SSE path.** `stream-pod-logs` / `stream-worker-logs` build their own `HttpError` and returned a bare 429; they carry the same hint now.
- **Human-scale units.** "wait ~29 minutes (1724s)" instead of "~1724s".

`HttpError.status` / `.body` are unchanged for programmatic callers. Changeset rewritten to match.

<details>
<summary>Testing</summary>

- 489 tests / 0 fail (510 / 0 after a local merge with current main), tsc + eslint + prettier clean on the diff
- `tests/rate-limit.test.ts` pins each failure mode above, plus a 20k-case deterministic property test against a naive over-reading oracle: any wait the hint quotes must be ≥ what a plain read of the header shows, unless hedged "at least"
</details>

justinwlin (approved):

Issue #63, opened 2026-07-29
update-endpoint silently drops GPU-pool exclusion entries in gpuIds on partial updates


AshishGapat:
## Bug: `update-endpoint` silently drops GPU-pool exclusion entries in `gpuIds`, even when only unrelated fields are provided

**Tool:** `update-endpoint`

**Description**

The tool description states "Only provided fields change," implying a true partial update. In practice, calling `update-endpoint` with only an unrelated field set (e.g. `imageName`) also reset the endpoint's `gpuIds`/GPU-pool config, specifically dropping per-card exclusion entries (entries prefixed with `-`, e.g. `AMPERE_48,ADA_48_PRO,-NVIDIA RTX A6000,-NVIDIA L40` → `AMPERE_48,ADA_48_PRO`).

This was only detectable after the fact via the RunPod dashboard's **Releases** tab diff — `get-endpoint` never surfaces the exclusion entries in its GPU pool response (it only returns the bare pool list), so there is no way to read-verify the full live `gpuIds` config before or after an update through the MCP tool alone.

**Steps to reproduce**

1. Create/have a Serverless endpoint whose GPU pool config includes at least one `-<GPU type>` exclusion entry (set via the RunPod dashboard or REST API).
2. Call `update-endpoint` with only an unrelated field, e.g. `{ "endpointId": "...", "imageName": "..." }`.
3. Check the endpoint's config in the RunPod dashboard's Releases tab.

**Expected:** Only `imageName` changes; `gpuIds` (including exclusion entries) is untouched.

**Actual:** `imageName` changes as expected, but the `gpuIds` exclusion entries are also dropped.

**Suspected root cause**

`update-endpoint` appears to reconstruct a full mutation payload (rather than sending a true partial patch), likely built from a prior `get-endpoint`-shaped read. Since that read shape doesn't include GPU-pool exclusion entries, any field not explicitly round-tripped gets reset to a default on write — not just the field(s) actually passed to the tool call.

**Suggested fix**

Either send a true partial patch (only literally the fields provided in the tool call), or have `get-endpoint`/the update path fully round-trip GPU-pool exclusion entries so they aren't silently lost.

---
*Issue drafted with AI assistance (Claude Code) based on an observed production incident; reproduction steps above are generalized/minimal, not copied from the original project.*

justinwlin:
investigating, thx for the flag

PR #69, opened 2026-07-30
fix(install): Claude Code not detected or registrable on Windows (#56)

### LINKED ISSUE #56: Bug: Install wizard never detects Claude Code on Windows, even when it's installed and working
### Summary

The `add`/`remove` install wizard shows Claude Code as not detected on Windows, even when it's correctly installed globally and `claude --version` works fine.

Image

**Root cause:** `findClaudeBinary()` in `src/install/clients.ts` falls back to `execSync('command -v claude')`. `command -v` is a POSIX shell builtin. On Windows, `execSync` runs through `cmd.exe` by default, which has no such builtin, so the call always throws regardless of whether Claude Code is installed. The hardcoded candidate paths checked before that (`~/.claude/local/claude`, `/usr/local/bin/claude`, `/opt/homebrew/bin/claude`) are also POSIX-only and never resolve on Windows.

Net effect: Claude Code cannot be auto-detected on a standard Windows setup running the installer from `cmd.exe` or `PowerShell`, even with a valid PATH installation.

### Environment

- **Operating System**: Microsoft Windows 11
- **Node.js**: `v22.14.0`
- **NPM**: `10.9.2`
- **Claude Code**: `2.1.212` (installed globally via `npm install -g @anthropic-ai/claude-code`)

### Steps to reproduce

```bash
npm install -g @anthropic-ai/claude-code
claude --version            # works fine Returns "2.1.212 (Claude Code)"
node dist/stdio.mjs add     # Claude Code shows as NOT detected
```

### Root cause (source)

```typescript
// src/install/clients.ts
function findClaudeBinary(): string | null {
  const candidates = [
    path.join(os.homedir(), '.claude', 'local', 'claude'),
    '/usr/local/bin/claude',
    '/opt/homebrew/bin/claude',
  ];
  for (const candidate of candidates) {
    if (exists(candidate)) return candidate;
  }
  try {
    execSync('command -v claude', { stdio: 'pipe' });
    return 'claude';
  } catch {
    return null;
  }
}
```

`execSync()` on Windows runs via `cmd.exe` unless another shell is explicitly configured. Since `cmd.exe` has no `command` builtin, `command -v claude` exits non-zero unconditionally - independent of whether `claude` is on PATH.

### Evidence

```powershell
> cmd /c "command -v claude"
'command' is not recognized as an internal or external command,
operable program or batch file.

> cmd /c "where.exe claude"
C:\Users\<user>\AppData\Roaming\npm\claude
C:\Users\<user>\AppData\Roaming\npm\claude.cmd
```

`where.exe` is the native Windows equivalent and resolves correctly. Reproduced identically via a standalone `execSync` script (not just interactive shell) and confirmed via instrumentation of `findClaudeBinary()` itself - details below.

<details>
<summary>Full reproduction details (instrumentation logs, standalone Node script, patched-build verification)</summary>

**Node reproduction (`execSync`, not shell):**
**Node reproduction (`execSync`, not shell):**
```
--- TEST 1: execSync('command -v claude') ---
FAILED
error.status / exitCode: 1
error.message: Command failed: command -v claude
'command' is not recognized as an internal or external command,
operable program or batch file.

stderr: 'command' is not recognized as an internal or external command,
operable program or batch file.

--- TEST 2: execSync('where.exe claude') ---
SUCCESS
stdout: C:\Users\<user>\AppData\Roaming\npm\claude
C:\Users\<user>\AppData\Roaming\npm\claude.cmd
```

**Instrumented `findClaudeBinary()` output:**
```
[INSTRUMENTATION] Checking candidate path: "C:\Users\91637\.claude\local\claude" | exists: false
[INSTRUMENTATION] Checking candidate path: "/usr/local/bin/claude" | exists: false
[INSTRUMENTATION] Checking candidate path: "/opt/homebrew/bin/claude" | exists: false
[INSTRUMENTATION] All candidates failed. Executing fallback 'command -v claude'...
[INSTRUMENTATION] Fallback FAILED (error: Command failed: command -v claude
'command' is not recognized as an internal or external command,
operable program or batch file.
). Returning: null
```

**After a temporary patch** (branching to `where.exe claude` on `win32`) and rebuild, Claude Code is immediately detected with no other behavior change:
```
*  Which agents should we set up?
|  [+] Claude Code (detected)
|  [+] Cursor (detected)
|  [+] Windsurf (detected)
|  [  ] Claude Desktop
|  [+] Visual Studio Code (detected)
```

</details>

### Why this matters

Claude Code is listed as a supported client in the README, and Windows is already handled elsewhere in this file (e.g. `claudeDesktopConfigPath()` branches on `win32`). Claude Code is the only client here whose detection relies on shell-based lookup rather than config-file existence, which is why it's uniquely affected.

### Suggested fix

Use a Windows-compatible lookup on `win32` (e.g. `where.exe`), or reuse an existing cross-platform executable-discovery helper if one exists in the codebase

### Regression risk

Extremely low - only the Windows branch of `findClaudeBinary()` changes. macOS/Linux behavior is untouched and this code only runs in the interactive installer, not at MCP runtime.

---

Happy to open a PR with this fix plus a regression test covering both platform branches - let me know if there's a preferred location for installer tests, since I didn't see existing coverage for `clients.ts`.

justinwlin:
Split from #67 — the install half. The GPU-exclusion half (#63) is in #70.

Closes #56

## Problem

Claude Code detection used the POSIX-only `command -v` fallback. On Windows, a standard
global npm installation exposes `claude.cmd`, so the wizard reported Claude Code as not
detected and could not register it.

## Fix

- Detect the documented native Windows install at
  `%USERPROFILE%\.local\bin\claude.exe` and the standard global npm install at
  `%USERPROFILE%\AppData\Roaming\npm\claude.cmd`.
- Do not search the current directory or project-local `PATH` on Windows. Candidates are
  canonicalized and confined to their expected install roots, so a symlink or junction
  cannot redirect credential-bearing execution into the current project.
- Never execute the npm `.cmd` shim. The wizard reads the installed package's
  `bin.claude` manifest entry, confines it to the canonical package root, and directly
  invokes the published native `bin\claude.exe` (or a legacy JavaScript target through
  the already-running Node executable), preserving argv without `cmd.exe`, `COMSPEC`,
  `PATHEXT`, or shell quoting.
- Sanitize `PATH`, `NODE_OPTIONS`, and `NODE_PATH` before invoking Claude Code.
- Keep POSIX discovery, including native installer locations and a guarded `command -v`
  fallback, without trusting project-local executables. The project-boundary check never
  treats `$HOME` itself as a project: a stray `package.json` or dotfiles `.git` in the
  home directory must not reject the standard install locations beneath it.
- Verify Claude Code add/remove against its user config. Existing malformed config is
  left unchanged and reported as a failure rather than being rewritten or reported as
  success.
- When Claude Code is not detected (custom npm prefix, redirected profile), the wizard
  prints the exact `claude mcp` command to register or remove the entry manually.
- Add Windows CI and tests for the real npm package layout, paths containing spaces,
  exact argv, poisoned environment variables, project shims, junction escapes, missing
  entrypoints, command failures, malformed config cleanup, and project markers in the
  home directory itself.
- Extend `pnpm type-check` to compile the test suite.

## Scope

Automatic Windows discovery intentionally covers the documented native path and the
standard npm prefix. Custom npm prefixes and redirected roaming profiles are not
auto-detected; the wizard's not-found message hands the user the manual command, and the
README documents manual setup for every client.

## Verification

- TypeScript source and test configurations: pass
- ESLint: pass
- Tests: 547 total, 539 pass, 0 fail, 8 skipped
- Production CJS/ESM/DTS build: pass
- Changed-file formatting and `git diff --check`: pass


lukepiette (changes requested):
Reviewed the full diff, the pre-PR code on `main`, and the published `@anthropic-ai/claude-code` package shapes. The Windows detection work, the add/remove verification against `.claude.json`, and the per-client config-dialect handling are unusually rigorous and consistently fail in the safe direction, and the test coverage is excellent. Requesting changes for one confirmed macOS/Linux regression, plus a second smaller one that should go in the same pass.

## Blocker: a project marker in `$HOME` disables detection of every Claude install under the home directory

`trustedClaudeBinary` rejects any candidate that resolves inside the outermost project boundary containing the cwd, and `projectBoundary()` walks from cwd all the way to the filesystem root, keeping the **outermost** ancestor containing any of the `PROJECT_BOUNDARY_MARKERS` (`.git`, `package.json`, lockfiles, ...).

If `$HOME` itself contains any of those markers (a dotfiles `.git`, or a stray `package.json` from an accidental `npm init` in home, both common), then for any cwd under `$HOME` the boundary becomes `$HOME`, and the wizard rejects:

- `~/.local/bin/claude`: the default native-installer location this PR itself adds as the first candidate
- `~/.claude/local/claude`: the migrated-installer location
- every nvm/asdf/Volta/fnm install found via the `command -v` fallback

Only `/usr/local/bin/claude` and `/opt/homebrew/bin/claude` survive. All of these worked on `main`, so this is a detection regression in a PR whose purpose is detection. The changeset covers the launched-from-`$HOME` case, but only for an *unmarked* home; the tests never model a marker in an ancestor of the home directory (every test home is a sibling tmpdir), so the suite cannot catch it.

Possible fixes: cap the `projectBoundary` walk at the canonical homedir, exclude homedir and its ancestors from being a boundary, or never reject a candidate that resolves inside one of the explicit trusted roots (`~/.local/bin`, `~/.claude/local`).

## Should fix in the same pass: `os.userInfo()` failure aborts all detection

In `findClaudeBinary`, when `os.userInfo()` throws (process UID with no passwd entry: OpenShift/arbitrary-UID containers, some CI sandboxes), the catch does `return null`, skipping even the POSIX `command -v` fallback, which needs no homedir at all. On `main`, `os.homedir()` never throws, so those environments regress from working to undetectable. The catch should fall through to the PATH lookup instead of returning.

## Non-blocking notes

- **Windows scope**: custom npm prefixes and corporate-redirected `APPDATA` (folder redirection/OneDrive) are skipped by design, and that population is arguably the one most likely to have hit #56. A more specific failure message than "claude CLI not found" would help those users self-diagnose.
- **Forward-compat coupling**: `interpretAddResult` turns a successful `claude mcp add` into a hard "Nothing was configured" failure when the entry is not at top-level `mcpServers` of `~/.claude.json` afterwards. Exact for claude 2.1.x, but if a future Claude Code moves user-scope storage, every add becomes a false failure. Worth a version-drift comment.
- **POSIX child PATH sanitization** (`sanitizedClaudeEnv`) could break `claude` wrapper scripts that rely on `~/bin` or Linuxbrew paths. Plausible only, not confirmed.
- **`ci.yml` is a full-file rewrite** (line-ending/quote churn) that obscures the two real changes (Windows matrix leg, Build skipped on Windows). Cosmetic, but it hurts blame.

## Verified fine while checking for breakage

- npm-global installs are not caught by the `node_modules/.bin` rejection: realpath resolves through `.bin` to the package target before the regex runs. Homebrew symlinks canonicalize inside their trusted roots.
- The Windows manifest logic matches every published package shape (`cli.mjs` at 0.2.9+, `cli.js` through 2.1.112, `bin/claude.exe` at 2.1.113+), and unknown shapes fail safe with a refused result and a copy-pasteable redacted command.
- No API breakage: the only importer (`wizard.ts`) uses an unchanged surface; the new exports are test-only.
- The edits to `credential-check.test.ts`, `handlers.test.ts`, and `pagination.test.ts` exist solely because `type-check` now compiles `tests/`; all type-level, nothing weakened. All 8 skipped tests are platform guards that execute on the other CI leg.
- `CLAUDE_CONFIG_DIR` semantics, the BOM handling, the duplicate-key read-back, and the 0600 create mode all check out.

With the boundary walk capped at (or excluding) the canonical home directory and the `userInfo` catch falling through to the PATH lookup, this is approve-with-comments territory. Nice work overall; the failure-direction discipline in this PR is genuinely rare.

justinwlin:
Thanks for the thorough pass — the review raced a push by a few minutes, so part of it was already addressed; the rest is now in.

**Blocker ($HOME project marker):** fixed in 2a9bd70 (pushed just before the review landed). `projectBoundary()` now stops at the canonical home directory without counting it — your suggested fix #1 — while still keeping the *outermost* marked ancestor below home, so a repo-root tool in a monorepo stays untrusted. Four regression tests: marked-`$HOME` detection (from home and from an unmarked subdir), project-under-marked-home symlink still rejected, nvm-style PATH hit from a marked home, and a Windows leg for a marked `%USERPROFILE%`.

**`os.userInfo()` failure:** fixed in 8bf2ee4. The catch now only disables the home-anchored candidates and falls through to the `command -v` lookup (uncapped boundary — the stricter direction). Deliberately not `os.homedir()`: its answer is the env-redirectable one `userInfo()` exists to avoid. `homedir: null` is the injectable sentinel, so the branch is covered by a test rather than an `os.userInfo` mock.

**Non-blocking:**
- Not-found message: now prints the exact `claude mcp add`/`remove` command (key redacted) for custom prefixes and redirected profiles, and the README documents manual setup per client.
- Version coupling: documented at the `.claude.json` reader — if a future CLI moves user-scope storage, that comment names the symptom (every add reports "Nothing was configured") and where to look.
- PATH sanitization + ci.yml churn: agreed both are plausible/cosmetic; leaving as-is unless you feel strongly. The ci.yml rewrite is line-ending normalization (CRLF→LF) — the real diff is the matrix `include`, `fail-fast: false`, and the Windows build skip.

CI is green on all four legs including windows-latest (548 tests).

justinwlin:
Both requested changes are done and pushed: the $HOME project-boundary blocker (2a9bd70) and the `os.userInfo()` fall-through to the PATH lookup (8bf2ee4), each with regression tests — details in the comment above. CI is green on all four legs including windows-latest (548 tests). Ready for re-review.

lukepiette (approved):

PR #74, opened 2026-07-31
feat: point agents at the OpenAPI and GraphQL specs via server instructions

lukepiette:
## Why

The MCP tools are curated projections of the Runpod APIs. When a tool doesn't expose a field or parameter an agent needs (e.g. `list-endpoint-workers` omits worker host machine info that GraphQL has), the agent dead-ends with no way to discover the wider surface. Found while dogfooding the server with Claude Code — the agent spent a full detour on deprecated `rest.runpod.io/v1` because nothing in the MCP surface mentioned v2 exists.

## What

- Adds `instructions` to the `initialize` response (most MCP clients prepend this to model context), linking the machine-readable API contracts: the REST v2 OpenAPI spec (`api.runpod.io/v2/openapi.json`) and the GraphQL schema reference (`graphql-spec.runpod.io`), with a note that prod GraphQL introspection is disabled.
- Moves `capabilities` into the `ServerOptions` argument of the `McpServer` constructor, where the SDK actually reads it (it was previously in the `Implementation` info object, where it is ignored).
- Changeset: minor.

## Test plan

- `pnpm build` clean, `pnpm lint` clean, `pnpm test` 502 pass / 0 fail.
- Manual stdio handshake against `dist/stdio.mjs`: `initialize` result now carries `instructions`; `serverInfo` unchanged (`Runpod API Server 3.1.0 [RUNPOD_REST_VERSION unset (default v2)]`).

Related Linear: CON-881 (SDL publishing), CON-880 (exec endpoint) — same dogfooding session.


lukepiette:
Post-push review (agent-verified against the built server over stdio and in-memory transports) caught one regression in the original diff, fixed in 1f4c5ea:

- **Dropped `resources: {}` from the advertised capabilities.** Once `capabilities` moved to `ServerOptions` (where the SDK actually reads it), the server began advertising a resources capability it doesn't implement — clients that enumerate resources on connect would get `-32601` on every session. The server registers tools only, so it now advertises tools only.
- **Added `tests/server.test.ts`**: an `InMemoryTransport` handshake test pinning the wire behavior — `serverInfo` carries no stray `capabilities` (the pre-existing bug this PR fixes), `resources` is not advertised, and `instructions` round-trips.

Also verified empirically: `api.runpod.io/v2/openapi.json` returns valid OpenAPI 3.1 (157 KB), `graphql-spec.runpod.io` is live, and prod GraphQL introspection returns Apollo `INTROSPECTION_DISABLED` — the instructions' claims hold. Full suite: 515 pass / 0 fail.

justinwlin (approved):
Approving. Small, correct, and the capabilities move is a real fix rather than a cosmetic one: `Implementation` is passthrough, so the old `capabilities` key rode along inside `serverInfo` on every `initialize` while the SDK ignored it. The new test pins that.

I re-verified every claim the instructions make against the live hosts:

| Claim | Result |
| --- | --- |
| `api.runpod.io/v2/openapi.json` | 200, OpenAPI 3.1, 29 paths, `servers: [api.runpod.io]` — byte-identical to `v2-rest.runpod.io/v2/openapi.json` |
| `graphql-spec.runpod.io` | 200 |
| prod introspection disabled | Apollo `INTROSPECTION_DISABLED` |
| REST v1 deprecated | matches the public docs warning in `api-reference/overview.mdx` |
| no resources implemented | zero `registerResource` / resource `setRequestHandler` in `src/`, so dropping `resources: {}` is right |
| both transports covered | `src/stdio.ts` and `src/http.ts` both construct via `createServer` |

Non-blocking notes, roughly in order of how much they matter:

1. **Two hostnames for one v2 API.** The instructions send agents to `https://api.runpod.io/v2`, but the tools call `https://v2-rest.runpod.io/v2` (`src/_shared/backend.ts` default, `docs/configuration.md`). Same backend today — identical spec bytes, identical 401 body — but the agent-visible text names one host while tool traffic goes to another, and if they ever diverge the instructions point at the wrong one with nothing to catch it. Pre-existing and out of scope here; worth either aligning the default to the documented canonical base or adding a clause noting they are the same API.

2. **Instructions are silent on which REST version the tools are actually on.** `serverInfo.version` carries `[RUNPOD_REST_VERSION unset (default v2)]`, but hosted HTTP resolves `auto` to v1. Telling the agent "v1 is deprecated, prefer v2" while the tools it is calling may be on v1 is a mixed message; one clause pointing at the version tag would close it.

3. **The GraphQL example will rot.** "per-worker host machine details" and "no REST equivalent" are true today — the prod v2 `Worker` schema is `id, status, isStale, version, gpuCount, image, uptimeSeconds, gpuTypeId, dataCenterId, startedAt`, with no machine fields — but when v2 adds them the instructions become wrong and no test fails. Consider softening to "may expose data the REST API does not yet cover", or dropping the specific example.

4. **Key hunting.** "Bearer auth with a Runpod API key" invites an agent to go looking through env vars and config files for a credential. In hosted mode the key is the caller's Bearer token held server-side and the model never sees it, so that path dead-ends anyway. A short "use the key the client already provides; do not go looking for one" would help.

5. **Test nit:** `assert.equal('capabilities' in (serverInfo ?? {}), false)` passes vacuously if `serverInfo` is `undefined`. The assertion above it catches that, but an `assert.ok(serverInfo)` first is tighter.

6. `capabilities: { tools: {} }` is redundant — `registerTool` merges `{ tools: { listChanged: true } }` before connect — but explicit is fine, and the comment already covers the resources half.

One cost note for the record: this injects roughly 1.4 KB (~350 tokens) into every session's context. Cheap relative to the dead-end it removes.

lukepiette:
Addressed the review notes in 3e6db3d:

1. **Dual v2 hosts** — instructions now say `v2-rest.runpod.io/v2` serves the same API and the tools may call either host.
2. **REST version visibility** — added a pointer to the `serverInfo.version` tag (e.g. `[RUNPOD_REST_VERSION=v2]`) so agents can see which version the tools are actually on.
3. **GraphQL rot** — softened to "may expose data the REST API does not yet cover" and dropped the per-worker machine example.
4. **Key hunting** — added "use the key this client is already configured with — do not search the environment or filesystem for a credential."
5. **Vacuous assert** — `assert.ok(serverInfo)` now precedes the property assertions.
6. Left `capabilities: { tools: {} }` explicit per your note.

`pnpm build` / `lint` clean, 515 pass / 0 fail.

PR #75, opened 2026-08-04
feat: make api.runpod.io/v2 the default REST v2 host

lukepiette:
## Why

Follow-up to #74. The review there flagged (note 1) that the server instructions point agents at `api.runpod.io/v2` while the tools' default base is `v2-rest.runpod.io/v2` — same backend, two names, and nothing would catch a divergence. Confirmed with @justinwlin that we're good to converge on `api.runpod.io/v2`, the host the public docs and the OpenAPI `servers` field name.

## What

- **`src/_shared/backend.ts`**: `restV2Base` default is now `https://api.runpod.io/v2`. `RUNPOD_REST_V2_API_URL` still overrides, so pinned deployments (including dev via `v2-rest.runpod.dev/v2`) are unaffected. The hosted credential-gate in `src/http.ts` compares against `restV2Base({})`, so it follows the new default automatically.
- **`src/server.ts`**: instructions drop the dual-host clause — one canonical host, noted as the same one the tools call.
- **Tests**: default-URL assertions in `handlers.test.ts` / `backend.test.ts` updated; the credential-check "pinned to its own default value" case now pins the new default (preserving that test's intent).
- **Docs/scripts**: `docs/configuration.md` table and the `fetch-v2-spec.ts` usage comment name the new host. `CHANGELOG.md` and `test.md` left untouched (historical records).
- Changeset: minor.

## Verification

- Live parity re-checked before the swap: `api.runpod.io/v2/openapi.json` 200 (156,963 B), `openapi.yaml` 200 and byte-count-identical to the `v2-rest` copy (204,587 B), unauthenticated `/v2/pods` returns 401 as expected.
- `pnpm build` clean, `pnpm lint` clean, `pnpm test` 525 pass / 0 fail (8 pre-existing skips).


justinwlin (approved):
Approving. Verified the host swap independently rather than taking the parity claim on faith:

- `api.runpod.io/v2/openapi.json` and `v2-rest.runpod.io/v2/openapi.json` both 200 and are **byte-identical** (156,963 B, `cmp` clean); unauthenticated `GET /v2/pods` returns 401 on both.
- No residual `v2-rest.runpod.io` anywhere in `src/`. The only remaining references are `CHANGELOG.md:106` and `test.md:489`, both historical records — correctly left alone.
- Confirmed the hosted credential gate follows the new default automatically: `src/http.ts:128` compares `restV2Base(env)` against `restV2Base({})`, so there is nothing to update there.
- Merges cleanly with #76 despite both touching `tests/handlers.test.ts`; the merged tree runs 527 pass / 0 fail / 8 skipped.

One non-blocking observation, since the note at `test.md:499` flags it as unverified. Pinning the **old** default host now reads as "REST moved" and silently switches the pre-flight off:

```
RUNPOD_REST_V2_API_URL=https://v2-rest.runpod.io/v2
  → restMoved = true, graphqlMoved = false
  → credentialCheckMayBeWrongEnvironment() = true → gate skipped
```

Confirmed empirically with this repo's own `gateRan()` harness: `gateRan({ RUNPOD_REST_V2_API_URL: 'https://v2-rest.runpod.io/v2' }) === false`.

**This does not affect production** — checked the Vercel env for this project and no host variable is pinned at all (`RUNPOD_REST_VERSION=v2` is the only RUNPOD_* var), so both sides stay on their defaults and the gate keeps running. It only matters for a self-hosted/stdio user who pinned that host back when it *was* the default, and who would lose the pre-flight with no signal.

Worth a small follow-up rather than a change here: treat `v2-rest.runpod.io/v2` as an alias of the default in `sameHost`, and add a `gateRan` case pinning it. The updated "pinned to its own default value" test covers the new host only, so that path is currently untested.

lukepiette:
Addressed the review observation in 764056e rather than deferring it: `normalizeUrl` now maps the retired `v2-rest.runpod.io` host onto `api.runpod.io` before comparison (prod pair only), so a deployment that pinned the old default keeps the credential pre-flight. Added the `gateRan` case pinning the old host, plus a negative case confirming `v2-rest.runpod.dev` still reads as a genuinely different environment and is not normalized into prod. Type-check, lint, and tests all clean (526 pass / 0 fail / 8 pre-existing skips).

PR #76, opened 2026-08-04
fix: guard v2-required pod name in create-pod

lukepiette:
## Why

Found while auditing every tool call against the live v2 OpenAPI spec (`api.runpod.io/v2/openapi.json`, follow-up to #74/#75): `CreatePodRequest` marks `name` **required**, but `create-pod`'s schema has it optional with no handler guard. A direct-image GPU create without a name sent a body the spec rejects — the caller got the API's raw 422 instead of a clean local error. `create-endpoint` already guards `name` for exactly this reason, and every other create tool requires it in its zod schema; `create-pod` was the one inconsistency.

## What

- **`src/tools/pods.ts`**: on the v2 GPU direct-image path, a missing `name` (with no `templateId`) now returns a local 400: "Provide a name for the pod (deploying from a templateId inherits the template name)." The guard deliberately does **not** apply where the spec doesn't:
  - **Template deploys** pass through — the template's `name` is merged in as the default (pinned by the existing template-merge test asserting `body.name === 'pytorch-template'`).
  - **CPU pods** pass through — they route to v1, which has no such requirement (new test).
- `name` param description updated so agents know when it's required.
- Four existing tests that exercised the v2 POST without a name now pass one; two new tests pin the guard and the CPU pass-through.
- Changeset: patch.

## Spec-audit context

This was the only inconsistency found. Everything else checked out: all 15 v2 paths/methods the tools construct exist in the spec; all query params (logs `source`/`tail`/`since`, billing windows/buckets/`lastN`, catalog `include=AVAILABILITY`) match including enums; all request bodies match field-for-field after resolving `$ref`/`allOf` (incl. the `additionalProperties: false` schemas); v1-only query filters are correctly never sent on v2; and the vendored parity fixture's operation set is identical to the live prod spec.

## Test plan

- `pnpm build` / `pnpm lint` clean, `pnpm test` 527 pass / 0 fail (8 pre-existing skips).

### DIFF
--- a/.changeset/create-pod-name-guard.md
+++ b/.changeset/create-pod-name-guard.md
@@ -1,0 +1,5 @@
+---
+'@runpod/mcp-server': patch
+---
+
+`create-pod` now returns a clean local 400 when `name` is omitted on a v2 GPU create without a `templateId`, matching the v2 spec (`CreatePodRequest` requires `name`) instead of surfacing the API's raw 422. Template deploys (name inherited from the template) and CPU pods (routed to v1) are unaffected.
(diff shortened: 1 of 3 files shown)


justinwlin (approved):
Approving. Guard placement and the two deliberate pass-throughs check out against the code, not just the description:

- The guard at `src/tools/pods.ts:257` sits after GPU/CPU disambiguation and before the CPU→v1 branch, so CPU pods reach v1 unaffected (pinned by the new test).
- Template deploys genuinely still work: `mapPodCreateToV2` runs its body through `compact()` (`src/_shared/mappers.ts:40`), which drops `undefined`, so the template's `name` survives the `{ ...podBodyFromTemplate(template), ...body }` spread. The existing `body.name === 'pytorch-template'` assertion is a real pin, not an incidental one.
- v1 mode is untouched — the guard is inside the `backend.version === 'v2'` block.
- 527 pass / 0 fail locally, including merged together with #75 (clean auto-merge on `tests/handlers.test.ts`).

Two nits, neither worth blocking on:

1. `compact()` keeps empty strings, so `name: ''` **with** a `templateId` overrides the template's name with `''` and still earns a raw 422. The guard treats `''` as missing on the direct-image path (`!params.name`) but the template path lets it through. Pre-existing, and arguably fine.
2. A `templateId` whose template carries no `name` still surfaces the raw 422. The spec requires template names, so this is close to unreachable — noting it only for completeness.

PR #77, opened 2026-08-04
feat: create-pod SSH access via sshPublicKey; fix v2 gpu.count default

### LINKED ISSUE #73: create-pod should expose SSH terminal access and inject account public keys
## Problem

The MCP `create-pod` tool does not expose an option equivalent to the Runpod console's **Enable SSH terminal access** setting or `runpodctl pod create --ssh` (which currently defaults to true).

This is especially problematic for users of the hosted MCP with OAuth who cannot generate/use a standalone API key for `runpodctl`.

## Observed behavior

- A Pod was created through the hosted Runpod MCP using OAuth and a user template.
- The MCP schema allowed exposing `22/tcp`, but offered no SSH enablement or public-key option.
- Pod creation returned a valid-looking direct TCP port mapping for port 22.
- Starting `sshd` in the template was not sufficient: the user's Runpod account SSH keys were not installed in `/root/.ssh/authorized_keys`, so the direct connection fell back to asking for a password and rejected the offered key.
- Runpod's Basic SSH proxy command worked for an interactive terminal, but that proxy requires a PTY and does not support SCP, SFTP, or rsync.

This makes a Pod appear SSH-ready when it is not usable for normal file-transfer workflows.

## Expected behavior

`create-pod` should expose a high-level SSH option comparable to the UI/CLI, for example:

```json
{
  "enableSsh": true,
  "sshPublicKey": "optional per-Pod override"
}
```

Suggested behavior:

1. Default `enableSsh` consistently with `runpodctl --ssh=true`, or make the default/limitation explicit.
2. Install the user's Runpod account public keys, or accept a per-Pod public-key override using the platform's documented SSH-key mechanism.
3. Ensure or clearly validate the requirements for full SSH: public IP support, `22/tcp`, an SSH daemon, and `authorized_keys`.
4. Return whether Basic SSH and Full SSH are actually ready, along with usable connection commands.
5. Warn rather than returning a seemingly valid port-22 mapping when authentication has not been configured.

The raw REST API may expose only lower-level primitives rather than an SSH boolean, but the MCP can provide the same convenience orchestration already exposed by `runpodctl`.

## Why this matters

Hosted MCP + OAuth is the only available automation route for some users. Without SSH enablement in the MCP, they must manually repair every Pod through the web or Basic SSH terminal, and cannot use rsync/SCP/SFTP until that repair is complete.

justinwlin:
Fixes #73.

## Problem

Two independent defects left `create-pod` unable to produce an SSH-usable Pod:

1. **No SSH option.** Pods created through the MCP look SSH-ready (valid port-22 mapping) but reject keys: the REST API has no SSH switch, so nothing installs the user's key into `authorized_keys`. The console and `runpodctl pod create --ssh` set GraphQL `startSsh`, which injects the account's registered keys as a `PUBLIC_KEY` env var that official images turn into a running sshd. MCP callers had no equivalent, so direct SSH, SCP, SFTP, and rsync all failed.
2. **v2 GPU creates without `gpuCount` always fail.** The v2 spec documents `gpu.count` as optional (server defaults to 1), but in practice omitting it makes the scheduler match zero machines — every create returns a misleading "no instances available" 400 (verified live on both v2 hosts; the identical body with `count: 1` succeeds).

## Changes

New optional `sshPublicKey` param on `create-pod` (works on v1, v2, CPU-passthrough, and template deploys):

- Merges the key into the `PUBLIC_KEY` env var and ensures `22/tcp` is exposed — the same convention `startSsh` uses, reproduced client-side.
- Applied to the final body **after** the template merge, so a template's ports/env are extended, not clobbered.
- An existing `PUBLIC_KEY` (caller's or template's) is appended to, never overwritten.
- A value containing `PRIVATE KEY` is rejected with a 400 before it can reach the Pod's env.
- Success replies carry an `_ssh` note stating what was set and what full SSH still depends on (image honoring the convention, public IP).

And the v2 mapper now always emits `gpu.count` (defaulting to 1) instead of trusting the documented server-side default.

When REST v2 ships its native `startSsh` field (merged in the API repo, not yet deployed), the tool can additionally pass that through; `sshPublicKey` stays useful as a per-Pod override.

## Test plan

- `pnpm test` — 541 pass, 0 fail; type-check + lint clean.
- New unit tests for `applySshPublicKey` (port dedupe incl. case/whitespace, `22/udp` not counted, env merge, append-to-existing `PUBLIC_KEY`, input not mutated, key trimmed) and for the `gpu.count` default.
- New handler goldens: v1 body (param stripped from the wire, env+ports injected), v2 mapped body, template deploy extends template ports/env, CPU v1 passthrough, private-key rejection fires no request.
- **Live E2E (both API versions)**: drove this branch's build over real stdio MCP with a throwaway keypair — `create-pod` with only `sshPublicKey` (no manual ports/env, no `gpuCount`) → Pod came up with `ports: ["22/tcp"]` and `PUBLIC_KEY` exactly equal to the passed key → real `ssh` login succeeded with `authorized_keys` containing exactly that key → SCP round-trip succeeded. Verified on the v1 path and on the default v2 path (which fails entirely without the `gpu.count` fix). Pods deleted after.

<details>
<summary>Example call</summary>

```json
{
  "name": "my-dev-pod",
  "imageName": "runpod/pytorch:1.0.2-cu1281-torch280-ubuntu2404",
  "gpuTypeIds": ["NVIDIA A40"],
  "sshPublicKey": "ssh-ed25519 AAAA... me@laptop"
}
```

Resulting body gains `"ports": ["22/tcp"]`, `"env": {"PUBLIC_KEY": "ssh-ed25519 AAAA... me@laptop"}`, and `"gpu": {"id": "NVIDIA A40", "count": 1}`.
</details>


lukepiette (approved):
Approving. Ran the branch's full suite (541 pass / 0 fail, tsc clean) and chased the two scariest scenarios to ground:

- **`gpu.count: 1` cannot downgrade a template's GPU count.** `podBodyFromTemplate` copies only name/image/args/disk/ports/env/mounts/registry — a v2 template body never carries a `gpu` key, and GPU compute is always caller-supplied, so the `{...template, ...body}` spread has nothing to clobber. Explicit `gpuCount` still flows through (`gpuCount ?? 1`); only the omitted case changes, and the vendored spec itself documents `count: default 1`, so this stays correct even if the scheduler-side default is ever fixed.
- **`sshPublicKey` never reaches the wire.** Destructured once at the top; the remainder feeds the v1 passthrough, `mapCreate`, and the template merge alike, with `'sshPublicKey' in body === false` asserted on all three paths. The `PUBLIC_KEY` merge appends with the newline separator authorized_keys requires, runs after the template merge (extends, not clobbers — tested), and the 22/tcp dedupe handles case/whitespace with `22/udp` correctly not counting. Branch is already rebased on #75/#76, so no conflicts.

Non-blocking suggestions:

1. A whitespace-only `sshPublicKey` (`" "`) is truthy: it exposes 22/tcp, sets `PUBLIC_KEY: ""`, and returns the "key applied" note. A trim-emptiness check before applying closes it.
2. The private-key guard catches every RFC 7468 PEM header but not PuTTY `.ppk` files (`PuTTY-User-Key-File-2` contains no literal "PRIVATE KEY") or a header-stripped base64 blob. Case-insensitive `/private[-\s]?key/i` plus a positive format check (`/^(ssh-|ecdsa-|sk-)/` on the trimmed value) would catch both and reject garbage input early.
3. Pre-existing, not this PR: `gpuCount: 0` with gpuTypeIds emits `count: 0` and would 422 against the spec's `minimum: 1` — a `z.number().int().min(1)` on the tool schema is a cheap follow-up.

PR #81, opened 2026-08-05
fix(tools): clamp long-poll budgets to 45s on the hosted HTTP transport

justinwlin:
## Description

Same failure shape as #79, but per tool call: two tools plan to wait longer than the platform allows. The hosted server dies at Vercel's 60s `maxDuration`, yet `runsync-endpoint` held the upstream request open for 90s by default (300s via `wait`) and `stream-job` polled for up to 5 minutes. For any job slower than ~55s the platform reaped the function before the tool's own graceful-timeout path could run — the caller got a bare 504, chunks collected so far were discarded, and the instance held its provisioned memory until then. `stream-job` on hosted could never succeed for such jobs.

## How

- `ToolRuntime` now exposes `transport` (it was already on `ToolContext`).
- On `http`, `runsync-endpoint` always sends `wait` capped at 45000ms — including when the caller omits it, since the upstream default (90s) also exceeds the deadline. A job that outlives the wait comes back as a job ID plus a non-terminal status to poll with `get-job-status`.
- The clamp is no longer silent. A clamped reply used to be byte-identical to an unclamped one, so an agent that asked for `wait: 300000` and got `IN_QUEUE` could only conclude the job had been stuck for five minutes — and cancel it, 45 seconds in. Replies now carry `waitClamped: {requestedMs, effectiveMs, reason}`, but only when the caller can actually be misled: hosted, the request genuinely exceeded the budget, and the job has not finished.
- On `http`, `stream-job` stops polling after 45s and returns the chunks collected so far with `pollingTimedOut: true` and a resume note. `/stream/{jobId}` drains what it hands out, so calling `stream-job` again continues where the last call stopped.
- `stream-job`'s http polls send `wait=1000`. An empty `/stream` blocks server-side for its default 10s, and the loop only rechecks its budget between polls — so a chunk-sparse job polled at ~11s cadence and exited at ~54s, right back against the reaper. Capping the upstream hold puts the worst case at ~46s.
- Tool descriptions are built per transport. Stating both budgets ("5 minutes, or 45s if hosted") guaranteed every reader a clause that was false for them with no way to tell which; each caller is now told the one budget that applies.
- stdio keeps the full budgets and bare poll URLs — no platform deadline there.

## Follow-up: per-request timeouts (separate PR)

**This PR does not make the 45s budget enforceable — it makes it correct.** The budget is checked *between* polls, and no outbound request in this server carries a client-side timeout: `RequestInitLike` in `src/_shared/http.ts` is `{method, headers, body?}` with no `signal`, and node-fetch has no default. So a single wedged connection — a worker that accepts the socket and goes quiet — never returns, the budget check below it never runs, and Vercel still reaps at 60s.

| Scenario | Before | This PR | With timeouts |
|---|---|---|---|
| Slow-but-responsive job | **always** reaped at 60s | ends ~46s | ends ~46s |
| One wedged connection | reaped at 60s | **still** reaped at 60s | ends at budget |

The left column is the bug that was actually costing money — it fired on every chunk-sparse job. The wedged-connection case is real but far rarer.

It is deliberately not fixed here because the fix belongs in the shared HTTP client that all ~30 tools use, not in the two job tools. Threading a signal touches `RequestInitLike`, `FetchLike`, the `HttpClient` signature, `serverlessRequest`, and both call sites; a mistake there breaks pods and endpoints, not just streaming. It also has a trap worth its own review: `runsync-endpoint` deliberately asks the server to hold a request for up to 45s (300s on stdio), so a naive 30s default would break it. Precedent for the pattern already exists in `src/_shared/credential-check.ts` and `src/_shared/backend.ts`, both of which use `AbortSignal.timeout` for exactly this reason.

Being worked on now against `origin/main`; will link here when it opens.

## Tests

- runsync http clamp: explicit wait above cap, omitted wait, wait under cap untouched, stdio passthrough.
- `waitClamped`: present when clamped + non-terminal (and for an omitted wait, against the 90s default); absent when the job COMPLETED, when the wait was under the cap, and on stdio.
- stream-job budget on both transports via a shared fake-clock helper — http stops at exactly 47 polls each carrying `wait=1000`, stdio runs to 302 with bare URLs. The stdio case is the only thing preventing a future "simplification" from clamping both.
- A test asserts `HTTP_LONG_POLL_BUDGET_MS` still fits inside `vercel.json`'s `maxDuration` with headroom for the pre-flight — that relationship is the premise of the clamp and nothing else linked the two files.
- Zod bounds for `runsync-endpoint.wait` (rejects 0/999/300001), following the existing pattern in `tools-registration.test.ts`. Handler tests bypass schema validation, so this bound needed its own coverage.
- Full suite: 544 pass / 0 fail on Node 18.20 and 24.2; type-check, lint, prettier, build clean.

<details>
<summary>Review passes — what they caught</summary>

Six review agents across two rounds. Beyond the `/stream` server-side block (the ~54s overshoot above, fixed in `4e05a4a`):

- **`stream-job`'s description actively misled.** It advertised "an unbudgeted stream, GET .../stream/{jobId}". `/stream` is a drain-poll — `LPopCount(…, 200)` then one `BLPop` — so a single GET returns only what is buffered, not the stream. Rewritten to say poll until terminal.
- **`wait: 0` reintroduced the bug.** `Math.min(0, 45000)` is `0`, which is falsy, so the query was dropped and the request inherited the upstream 90s default. Unreachable through Zod's `.min(1000)`, but handlers are called directly — as these tests do — and `catalog.ts` guards the same way for the same reason. Now tested against `undefined`.
- **A budget *increase* hung the suite instead of failing it.** Mutation-tested: `node --test` has no default timeout, so the drive loop's `await pending` never settled and the file's remaining ~90 tests never ran. Now fails in 1s with a legible message instead of hanging 114s.
- **The fake clock hardcoded its tick** rather than importing the real poll interval, so halving the interval would have doubled the live request rate with tests still green.
- Corrected: the clamped reply returns whatever status the job is in — usually `IN_QUEUE` for the slow endpoints this exists for, not `IN_PROGRESS`.

`?wait=1000` on `/stream` is verified live against production, not just against the service source. Controls: `wait=999` and `wait=300001` both return `400 "wait is invalid; must be between 1000 and 300000"`, while `wait=1000` passes validation and behaves identically to sending no parameter at all.

</details>


src/tools/jobs.ts
```diff
@@ -222,7 +232,19 @@ export function registerJobTools(server: McpServer, rt: ToolRuntime): void {
...
+      // both exceed the platform deadline, so the wait is ALWAYS sent there,
+      // clamped — an omitted wait would otherwise hold the function past 60s.
+      // When the clamped wait expires upstream, the response carries the job
+      // ID + IN_PROGRESS, which is exactly the documented poll-with-
+      // get-job-status path. stdio passes the caller's wait through untouched.
+      const effectiveWait =
+        rt.transport === 'http'
+          ? Math.min(wait ?? 90_000, HTTP_LONG_POLL_BUDGET_MS)
+          : wait;
+      const path = effectiveWait
```
donovanclarke: nit: base is the same. why not append the whole `?wait=...`?

donovanclarke (approved):
lgtm, 

note: comment length is pretty verbose. Could be worthwhile to make them more concise :) (non-blocker though)

PR #85, opened 2026-08-18
feat: expose v2 CUDA constraints (gpu.allowedCudaVersions / gpu.minCudaVersion)

brodykellish:
## What rphttp2 2.9.0 changed on the wire

The CUDA host constraints (`allowedCudaVersions`, `minCudaVersion`) moved from the top level of pod/endpoint bodies into the `gpu` block (pre-2.9.0 servers 422 the nested spelling as an unknown property; 2.9.0 servers 422 the old top-level spelling). Serverless gained a `gpu.minCudaVersion` floor. On `PATCH /v2/serverless`, `gpu.pools` became optional (a CUDA-only patch need not resend the pool list), and sending `pools` now replaces the GPU selection wholesale — clearing `excludedTypes`.

The MCP server needed nothing to *survive* 2.9.0 (it never sent or read the affected fields over REST); this PR *adopts* the new schema.

## What this adds

- `allowedCudaVersions` / `minCudaVersion` params on **create-pod**, **create-endpoint**, and **update-endpoint**, emitted under `gpu.*`. Client-side validation names the failure instead of relaying a raw 422/400: major.minor grammar (a bare `"12"` is rejected with the fix in the message), set-XOR-floor exclusivity, GPU-pods-only. Update supports the spec's clear sentinels (`[]` clears the set, `""` clears the floor).
- **Pools-less `gpu` PATCH**: a CUDA-only (or count-only — previously silently dropped) endpoint update no longer resends `gpuPoolIds`.
- **Exclusion-wipe warning**: when `update-endpoint` sends `gpuPoolIds` and the endpoint carries `excludedTypes` (set via console or `set-endpoint-gpus`), the reply carries a `_warning` naming the cleared exclusions.
- **list-gpu-types** gains the spec's `minCudaVersion` availability filter (bare major legal here — it only widens a read).
- **Vendored spec refreshed**; the parity gate flagged nine new unmapped operations (account SSH keys, Instant Clusters, public template catalog) — allowlisted as explicit to-dos, not silently.
- v1 mode rejects the new params with a clean 501 pointing at `RUNPOD_REST_VERSION=v2` (the `templateId` precedent).

**Deploy gate — cleared.** This needed prod `api.runpod.io` on rphttp2 2.9.0, since pre-2.9.0 servers 422 the gpu-nested fields as unknown properties. Verified against the live production spec (`api.runpod.io/v2/openapi.json`, Runpod REST API 2.0.0): `allowedCudaVersions` and `minCudaVersion` are present, and the semantics this PR codes against are the deployed ones — the non-empty XOR rule, "an explicit `[]` states no constraint", and numeric (not decimal) version comparison. Nothing blocks release.

28 new tests (14 mapper, 14 handler); suite 737 tests, 729 pass / 8 pre-existing env-gated skips; tsc + eslint clean, prettier clean on touched files (three files carry pre-existing formatting warnings from `main`, which CI does not gate).

## Review follow-ups

All review comments are resolved. Copilot's three inline comments were fixed in `b4da9a7` (misleading `get-capacity` product reference, `gpuPoolIds: []` silently no-opping on update, and the note explaining v2-only fields on `V1PodParams`). A later commit addressed Donovan's four nits and the two comments Copilot suppressed:

- **catalog**: the `list-gpu-types` query is built with `URLSearchParams` instead of nested template ternaries. All three URL shapes are pinned by exact-equality assertions, so the output is byte-identical.
- **endpoints**: the scaler-type narrowing is named rather than repeating the literal to mean "or the default", and documents that an unknown future scaler type fails closed onto the stricter bound. No const-object indirection: the values are already a TS union and a zod enum, so it would add a layer without adding safety.
- **mappers**: the two nested v2 wire objects are typed (`V2GpuConfig`, `V2Workers`) instead of `Record<string, unknown>`, so the shape is compile-checked and not only fixture-asserted. Needs `compact()`'s constraint widened to `T extends object`, since an interface has no index signature.
- **spec-parity**: dropped the stale "Empty today — every operation maps to a registered tool" paragraph, which contradicted the ten-entry allowlist directly beneath it.
- **update-endpoint**: the v2 mutable-fields list omitted `gpuPoolIds`/`gpuCount` and the CUDA constraints this PR adds.

`main` has been merged in, so the branch carries the CDN discovery-cache change from #86.

### DIFF
--- a/src/_shared/mappers.ts
+++ b/src/_shared/mappers.ts
@@ -33,6 +33,8 @@ interface V1PodParams {
   env?: Record<string, string>;
   dataCenterIds?: string[];
   containerRegistryAuthId?: string;
+  // v2-only: mapped into gpu.allowedCudaVersions / gpu.minCudaVersion; the v1
+  // path never sends them.
   allowedCudaVersions?: string[];
   minCudaVersion?: string;
 }
--- a/src/_shared/mappers.ts
+++ b/src/_shared/mappers.ts
@@ -33,6 +33,8 @@ interface V1PodParams {
   env?: Record<string, string>;
   dataCenterIds?: string[];
   containerRegistryAuthId?: string;
+  allowedCudaVersions?: string[];
+  minCudaVersion?: string;
 }
 
 // Drop undefined entries so we never emit explicit `undefined`/`null` the API
@@ -93,11 +95,20 @@ function containerConfigToV2(p: {
 // always emit the documented default explicitly.
 function gpuConfigToV2(
   gpuTypeIds?: string[],
-  gpuCount?: number
+  gpuCount?: number,
+  allowedCudaVersions?: string[],
+  minCudaVersion?: string
 ): Record<string, unknown> | undefined {
   const id = gpuTypeIds?.[0];
   if (id === undefined) return undefined;
-  return { id, count: gpuCount ?? 1 };
+  // CUDA host constraints are gpu-nested since 2.9.0 (never body top-level).
+  // Passed through as given: an explicit [] is a spec-legal "no constraint".
+  return compact({
+    id,
+    count: gpuCount ?? 1,
+    allowedCudaVersions,
+    minCudaVersion,
+  });
 }
 
 export function mapPodCreateToV2(params: V1PodParams): Record<string, unknown> {
@@ -124,7 +135,12 @@ export function mapPodCreateToV2(params: V1PodParams): Record<string, unknown> {
     dataCenterIds: params.dataCenterIds?.length
       ? params.dataCenterIds
       : undefined,
-    gpu: gpuConfigToV2(params.gpuTypeIds, params.gpuCount),
+    gpu: gpuConfigToV2(
+      params.gpuTypeIds,
+      params.gpuCount,
+      params.allowedCudaVersions,
+      params.minCudaVersion
+    ),
   });
 }
 
@@ -169,6 +185,8 @@ interface V2EndpointParams {
   ports?: string[];
   env?: Record<string, string>;
   containerRegistryAuthId?: string;
+  allowedCudaVersions?: string[];
+  minCudaVersion?: string;
 }
 
 type EndpointType = 'QUEUE' | 'LOAD_BALANCER';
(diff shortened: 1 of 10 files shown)


src/tools/catalog.ts
```diff
@@ -76,9 +89,35 @@ export function registerCatalogTools(server: McpServer, rt: ToolRuntime): void {
...
         const raw = await callRestUrl(
           `${backend.base}${backend.list}${
-            wantAvailability ? `?include=AVAILABILITY&product=${product}` : ''
+            wantAvailability
+              ? `?include=AVAILABILITY&product=${product}${
+                  params.minCudaVersion !== undefined
+                    ? `&minCudaVersion=${encodeURIComponent(params.minCudaVersion)}`
+                    : ''
+                }`
+              : ''
```
donovanclarke: nit: can we simplify this nested ternary for readability?

src/tools/endpoints.ts
```diff
@@ -411,24 +451,69 @@ export function registerEndpointTools(
...
           | undefined;
-        scalerType =
-          current?.scaling?.type === 'REQUEST_COUNT'
-            ? 'REQUEST_COUNT'
-            : 'QUEUE_DELAY';
+        if (needScalerRead) {
+          scalerType =
+            current?.scaling?.type === 'REQUEST_COUNT'
+              ? 'REQUEST_COUNT'
+              : 'QUEUE_DELAY';
```
donovanclarke: nit: minor concern around usage of magic strings. Seems like there isnt a project convention, but ideally could be simplified by usaing a const as object pattern?

src/_shared/mappers.ts
```diff
@@ -202,14 +222,24 @@ function endpointScaling(
...
+// handler's guard, not the API, reports the omission — CUDA/count fields never
+// ride without a pool list on create. UPDATE (2.9.0) makes every gpu field
+// optional, so a pools-less gpu (CUDA-only or count-only patch) is emitted as
+// given; an empty object is still dropped. CUDA clear sentinels pass through:
+// [] clears the set, "" clears the floor.
 function endpointGpuConfig(
-  pools?: string[],
-  count?: number
+  params: V2EndpointParams,
+  mode: 'create' | 'update'
```
donovanclarke, line 252: note: comment around typing etc below.

src/_shared/mappers.ts
```diff
@@ -227,7 +257,10 @@ function endpointWorkers(
 
 // The half of the body shared by create and update: container config, compute,
 // placement and workers. Only `type` and `scaling` differ between the two.
-function endpointCommonToV2(params: V2EndpointParams): Record<string, unknown> {
+function endpointCommonToV2(
+  params: V2EndpointParams,
+  mode: 'create' | 'update'
```
donovanclarke, line 281: here as well :)

justinwlin (approved):

PR #87, opened 2026-09-02
[Reference] REST v2 spec-generated MCP server with the published SDK

justinwlin:
Full reviewed reference for the REST v2 spec-generated MCP server. **Keep this PR as reference; merge the smaller stack below instead.**

| Order | PR | Review focus |
| --- | --- | --- |
| 1 | [#90](https://github.com/runpod/runpod-mcp/pull/90) | Generation and drift checks |
| 2 | [#91](https://github.com/runpod/runpod-mcp/pull/91) | SDK and API clients |
| 3 | [#92](https://github.com/runpod/runpod-mcp/pull/92) | Shared MCP surface |
| 4 | [#93](https://github.com/runpod/runpod-mcp/pull/93) | Hosted agent learning |

Merge **1 → 2 → 3 → 4**. Each PR targets the preceding layer. After squash merging, rebase/retarget the remaining stack before continuing. The final stack tree matches this branch exactly.

Review fixes include shared request deadlines, bounded job polling, per-caller worker diagnosis caches, sparse REST GPU updates, structural schema generation, and error/journal input handling. Release notes are consolidated by layer.

Validation: all 366 tests, type checks, lint, builds, generation checks and live REST spec comparison passed. Live REST reads and a disposable zero-worker GPU update passed; deletion was confirmed. Stalled transports were injected locally. Fresh packed ESM/CJS entrypoints initialize with all 68 tools and a bundled SDK. Production Convex storage was not deployed or exercised.

PR #90, opened 2026-09-14
[1/4] Generate REST v2 tool schemas and check spec drift

justinwlin:
Add the vendored REST v2 contract, pure schema generator, and drift checks. Preserve schema literals and parameter overrides. This layer keeps the current MCP server behavior.

| Order | PR | Review focus |
| --- | --- | --- |
| 1 **(this PR)** | [#90](https://github.com/runpod/runpod-mcp/pull/90) | Generation and drift checks |

Merge **1 → 2 → 3 → 4**. This PR targets `main`. After a squash merge, rebase/retarget the remaining stack before merging the next PR.

Validation: Type check, lint, build, generation check and live spec comparison passed; 744 tests passed (8 existing skips).


specgen/scripts/fix_spec.py
```diff
@@ -0,0 +1,542 @@
+#!/usr/bin/env python3
```
lukepiette, line 1: Orphaned: nothing in `package.json`, CI, or `pull-spec.ts` runs this, and the docstring's claim that `spec/openapi.yaml` is always this script's output isn't true (AWS account id and `hunter2` are still in the spec, and the `neutral-secret-examples` anchor doesn't exist so it raises `PatchError`). `test_fix_spec.py` fails 5/7 locally. Suggest dropping both files from this PR, or wiring them into `spec:pull` and fixing the tests.

specgen/generator-config.yaml
```diff
@@ -0,0 +1,86 @@
+# Controls the spec-to-tools generation. Every operation in spec/openapi.yaml
+# becomes a tool unless excluded here; the parity test enforces exactly that.
+
+# Operations excluded from generation, each with its reason and — where one
+# exists — the curated tool in src/tools/ that replaces it. The parity test
+# reads this mapping: an exclusion with no entry here fails it, and a
```
lukepiette, line 6: This parity test doesn't exist in this PR (no test reads `exclude` or `replacedBy`). Either land the assertion here or soften the comment until #92.

specgen/generator/tools.ts
```diff
@@ -0,0 +1,314 @@
...
+      name,
+      normalizeSchema(schema),
+    ])
+  );
+  const tools: GeneratedTool[] = [];
+  const names = new Set<string>();
+  for (const [path, pathItem] of Object.entries(spec.paths)) {
+    for (const method of HTTP_METHODS) {
+      const op = pathItem[method];
+      if (!op || Object.hasOwn(config.exclude ?? {}, op.operationId)) continue;
```
lukepiette, line 305: Consider tracking which `exclude` keys were actually matched and throwing after the loop if any were unseen. Otherwise an upstream operationId rename makes the excluded op silently reappear as a generated tool alongside the curated replacement.

specgen/generator/tools.ts
```diff
@@ -0,0 +1,314 @@
...
+  config: GeneratorConfig,
+  defs: Record<string, unknown>
+): GeneratedTool {
+  if (!op.operationId)
+    throw new Error(`Missing operationId: ${method} ${path}`);
+  const properties: Record<string, unknown> = {};
+  const required: string[] = [];
+  const params: GeneratedToolParam[] = [];
+  for (const param of operationParameters(pathItem, op, spec)) {
+    if (param.in !== 'path' && param.in !== 'query') continue;
```
lukepiette, line 229: Header and cookie params are silently dropped. Today the only header param (`Last-Event-ID`) is on excluded log ops so nothing is lost, but a future header param on a generated op would vanish without a signal. A `throw` matches the fail-loud style used elsewhere in this file.

.github/workflows/spec-drift.yml
```diff
@@ -0,0 +1,32 @@
...
+# API serves — the signal that `specgen/spec/openapi.yaml` needs a re-vendor
+# and `pnpm generate:tools` a re-run (see specgen/README.md).
+#
+# Runs daily, on demand, and on every PR — any PR shows at a glance whether
+# the vendored spec is current. It reports as its own check (spec-drift), so
+# whether a red result blocks merging is a branch-protection choice: leave it
+# out of the required checks and upstream churn stays visible without
+# stopping unrelated work.
+name: Spec drift
+
```
lukepiette, line 11: Nit: add `permissions:\n  contents: read` since this runs on every PR.

scripts/check-spec-drift.ts
```diff
@@ -0,0 +1,73 @@
...
+for (const id of moved)
+  console.error(
+    `  ~ ${id}  ${vendored.get(id)!.route} -> ${live.get(id)!.route} — its tool calls the OLD path`
+  );
+for (const id of changed)
+  console.error(
+    `  ~ ${id} — request body, parameters, or referenced schemas changed`
+  );
+console.error(
+  '\nFix: re-vendor the spec and regenerate (specgen/README.md), then run pnpm test.'
```
lukepiette, line 71: Nit: `specgen/README.md` doesn't exist until #92, so a red drift check on this branch points at a missing file.

lukepiette (commented):
Reviewed with a local checkout: `pnpm generate:check` passes, regenerating `tools.gen.ts` gives a zero diff, `pnpm spec:check` matches production (55 operations), type-check and lint are clean, and I spot-checked create-pod, update-endpoint, list-gpu-types, get-pod and pod-action against the vendored spec (paths, params, required-ness and `$defs` closure all match). Drift check has no false-pass path I could find.

One thing I'd like fixed before merge, plus a couple of small guards. Happy to approve once the first item is resolved.

1. **`specgen/scripts/fix_spec.py` and `test_fix_spec.py` are orphaned and broken.** Nothing references them (no package script, no CI step, `pull-spec.ts` writes the raw spec). The docstring says the vendored spec is always this script's output, but it isn't: the spec still has the concrete AWS account id and the `hunter2` example, the `neutral-secret-examples` patch raises `PatchError` against this spec, and 5 of 7 tests in `test_fix_spec.py` fail. It also references another repo's toolchain (`task generate`, `gopkg.in/yaml.v3`). Applying it would also break `spec:check`, since it mutates `components/schemas` which drift fingerprints. Suggest deleting both files here, or if you want the patch layer, wire it into `spec:pull` and make drift compare post-patch.
2. **Stale exclusions are silently ignored.** `generator-config.yaml` promises a parity test that fails on an exclusion without an entry, but no test reads `exclude`/`replacedBy` in this PR, and `tools.ts` only does a `hasOwn` check. If upstream renames an excluded operationId, the op quietly reappears as a generated tool next to the curated one. A one-line guard (throw if any `exclude` key was never seen) covers it.
3. Nits inline: header/cookie params are dropped without a signal, `spec-drift.yml` has no `permissions:` block, and the drift error text points at `specgen/README.md` which doesn't exist until #92.

justinwlin:
Addressed the review: removed the obsolete Python patch scripts, made generation reject stale exclusions and unsupported parameter locations, restricted the drift workflow to contents:read, and replaced the missing-doc link with runnable recovery commands. Added regression tests; generation still matches the vendored spec.

lukepiette (approved):
Re-reviewed 2a94e10. Everything from the first pass is addressed: the Python patch scripts are gone, stale exclusions and header/cookie params now fail generation with tests covering both, the drift workflow has `contents: read`, and the error text gives runnable commands. Re-ran `generate:check`, type-check and the suite locally: generation matches, 740 pass, same 6 machine-local `findClaudeBinary` failures as main.

One leftover nit, no action needed before merge: two comments (`spec-drift.yml:3`, `check-spec-drift.ts:6`) still point at `specgen/README.md`, which arrives in #92.

PR #91, opened 2026-09-14
[2/4] Add the published SDK and bounded API clients

justinwlin:
Add the published TypeScript SDK, per-caller contexts, deadline enforcement, runtime/log clients, and curated tool implementations. GPU selection uses a sparse REST PATCH. The existing server stays wired until part 3. Node.js 20+ is required.

| Order | PR | Review focus |
| --- | --- | --- |
| 2 **(this PR)** | [#91](https://github.com/runpod/runpod-mcp/pull/91) | SDK and API clients |

Merge **1 → 2 → 3 → 4**. This PR targets `review/specgen-01-generation`. After a squash merge, rebase/retarget the remaining stack before merging the next PR.

Validation: Type check, lint, build and generation check passed; 770 tests passed (8 existing skips). Live REST reads and a disposable zero-worker GPU update passed; cleanup was confirmed.


package.json
```diff
@@ -72,10 +73,11 @@
     "tsup": "^8.0.0",
     "tsx": "^4.7.0",
     "typescript": "^5.3.3",
-    "yaml": "^2.9.0"
+    "yaml": "^2.9.0",
+    "@runpod/typescript-api-sdk": "^0.1.0"
```
lukepiette, line 77: Since the SDK is bundled via `noExternal`, whatever the lockfile resolves is what ships to npm. A `^0.1.0` caret on a 0.x package means a `pnpm update` can silently change retry or deadline behaviour in a release with no diff in this repo. Suggest pinning exactly (`0.1.1`).

src/specgen/dispatch.ts
```diff
@@ -0,0 +1,155 @@
...
+// same hole until the 2026-09-09 release smoke: stream-pod-logs called with
+// `id` instead of `podId` fetched /pods/undefined/logs and answered 404
+// "pod not found" with a hint to re-verify the id. A wrong answer is worse
+// than an error, so name the key and list what the tool accepts.
+export function validateArguments(
+  inputSchema: { required?: unknown; properties?: unknown },
+  args: Record<string, unknown>
+): ToolResult | null {
+  const required = (inputSchema.required ?? []) as string[];
+  const missing = required.filter((name) => args[name] === undefined);
```
lukepiette, line 49: `null` passes this gate (`=== undefined` only), so `{ id: null }` reaches the SDK and becomes `/v2/pods/null`. Cheap to treat `null` as missing here too.

src/specgen/clients/runtime.ts
```diff
@@ -0,0 +1,74 @@
...
+  const apiKey = options.apiKey ?? process.env.RUNPOD_API_KEY;
+  const baseUrl = options.baseUrl || serverlessBase(process.env);
+  const fetchImpl = options.fetchImpl ?? fetch;
+
+  return async (endpointId, path, opts = {}) => {
+    if (!apiKey) throw missingKeyError();
+    const response = await boundedFetch(
+      fetchImpl,
+      opts.timeoutMs ?? DEFAULT_TIMEOUT_MS
+    )(`${baseUrl}/${endpointId}${path}`, {
```
lukepiette, line 44: Nit: `endpointId` is interpolated into the path unvalidated. Every current caller in `jobs.ts` gates it with `requireIds`, so nothing is exploitable today, but a check inside the client would protect future callers from `abc/../x` style ids.

lukepiette (approved):
Reviewed with a local checkout: type-check, lint and tsup build are clean; 764 tests pass (the 6 `findClaudeBinary` failures are a local-machine artifact, untouched by this PR, and green in CI).

Things I specifically traced and found correct:
- API key only ever lands in `Authorization` headers (context, runtime, sse, graphql); `ops.ts` logs a salted hash, never the token or args.
- Deadline fetch wraps the SDK retry layer so the full retry sequence fits inside the 30s budget and aborts body reads too. SSE byte cap cancels the stream. POST retries only on 429, so no double-create risk.
- All hosts come from `_shared/hosts.ts`; caller input never selects a host. GraphQL inlined values in `capacity.ts` are validated, `hub.ts` uses variables.
- `endpoint-gpus.ts` sends only the `gpu` sub-object and matches the spec's partial-update semantics, including clearing `minCudaVersion` alongside `allowedCudaVersions`.
- Polling and pagination are bounded (budget, consecutive-error cap, 4xx short-circuit, digits-only cursor, limit floored at 1).
- Smoke GPU script is gated behind `--allow-create` and cleans up in `finally`.

Approving. Two small things inline worth picking up, the SDK pin especially since the SDK is bundled into the published artifact.

justinwlin:
Addressed the review: pinned the bundled SDK to exactly 0.1.1, rejected null required arguments before dispatch, and validated endpoint IDs inside the runtime client. Added tests proving invalid inputs make no network requests.

PR #92, opened 2026-09-14
[3/4] Switch HTTP and stdio to the generated REST v2 surface

justinwlin:
Serve the same 68 tools and 10 skill resources on HTTP and stdio. Remove the legacy surface and v1 fallback, add hosted analytics and error recovery, and document renamed tools. Breaking: the package tools export is removed; old tool names map to the generated surface.

| Order | PR | Review focus |
| --- | --- | --- |
| 3 **(this PR)** | [#92](https://github.com/runpod/runpod-mcp/pull/92) | Shared MCP surface |

Merge **1 → 2 → 3 → 4**. This PR targets `review/specgen-02-sdk`. After a squash merge, rebase/retarget the remaining stack before merging the next PR.

Validation: Type check, lint, build, generation check and live spec comparison passed; all 334 tests passed. Fresh packed ESM/CJS entrypoints initialize and list 68 tools; the SDK is bundled.


.changeset/specgen-unified-surface.md
```diff
@@ -0,0 +1,5 @@
+---
+"@runpod/mcp-server": major
+---
+
+Serve the same REST v2 generated tools and skill resources over HTTP and stdio. Remove the legacy v1 fallback, hand-written tool surface, and package tools export. See specgen/old-mcp-tools.yaml for renamed tools and pod-action replacements. Include argument checks, compact lists, bounded log snapshots, upstream retry guidance, caller tracking, and opt-in hosted analytics. The REST SDK is bundled into both entrypoints.
```
lukepiette, line 5: `smithery.yaml`, `Dockerfile` and `.cursor/mcp.json` are deleted and the Smithery install section is gone from the README, but the changeset doesn't mention it. Worth one line so Smithery/Docker users get a migration note in the release, or keep `smithery.yaml` if the listing should survive.

specgen/old-mcp-tools.yaml
```diff
@@ -0,0 +1,70 @@
+# Parity manifest against the official Runpod MCP server
+# (github.com/runpod/runpod-mcp at f20d987, 54 tools). Every tool the official
+# server exposes maps to a tool this server exposes; tests/old-mcp-parity.test.ts
```
lukepiette, line 3: Nit: the check lives in `tests/specgen-gates.test.ts`, not `tests/old-mcp-parity.test.ts`, and curated tools are in `src/specgen/tools/` now.

.gitignore
```diff
@@ -143,3 +143,7 @@ vite.config.js.timestamp-*
 vite.config.ts.timestamp-*
 .vite/
 .vercel
+.env*
+
+# Vendored, locally built @runpod/sdk (see vendor/runpod-sdk) — npm publish pending
+!vendor/runpod-sdk/dist/
```
lukepiette, line 149: Nit: `vendor/runpod-sdk` doesn't exist on this branch (the SDK is bundled from npm via tsup `noExternal`). Stale leftover.

lukepiette (approved):
Reviewed with a local checkout: tsc, eslint, `generate:check` and a skills regen are all clean and in sync. Built `dist/stdio.mjs` with a dummy key: `tools/list` returns 68 and `resources/list` returns 10, and the HTTP path constructs the same server, so the surface is shared by construction. 328 tests pass (the 6 `findClaudeBinary` failures are the same local artifact as on main).

Parity: I extracted every `server.tool(...)` name from the deleted `src/tools/*.ts` on the base branch, 54 exactly, and each one has a served target in `specgen/old-mcp-tools.yaml`, enforced by `specgen-gates.test.ts`. Nothing silently disappears. Per-request context and transport are fresh per HTTP call and disposed on close, so no credential caching across callers. Analytics is a closed field set, HMAC'd distinct id, no-op without `POSTHOG_API_KEY`, fire-and-forget. LICENSE diff is just the brand casing. eslint only adds the convex ignore.

Approving. Three small things I'd fix before the major ships, inline: the install wizard still verifies keys against v1, the wrong-environment guard still reads the retired env spellings (which can silently disable the credential preflight), and the changeset doesn't mention that the Smithery and Docker distribution files were removed.

Two of those are in files this PR doesn't touch, so noting them here rather than inline:

- **`src/install/verify-key.ts:21`**: This still verifies against `rest.runpod.io/v1` via `RUNPOD_REST_API_URL`, while the rest of this PR retires v1 and `docs/configuration.md` says so. When v1 is decommissioned the wizard returns "check failed" for every valid key, and a dev deploy that sets only `RUNPOD_API_BASE_URL` verifies against prod. Suggest `sdkBase(env)` + `/v2/pods`, or the same GraphQL `myself` probe the hosted path uses.
- **`src/_shared/hosts.ts:16-22` / `src/http.ts:148-152`**: `RUNPOD_REST_API_URL` and `RUNPOD_REST_V2_API_URL` no longer configure any tool (docs say so), but `credentialCheckMayBeWrongEnvironment` in `src/http.ts` still reads them. A hosted deploy that carried over `RUNPOD_REST_V2_API_URL=<dev>` gets `restMoved=true` with `graphqlMoved=false`, so the credential preflight is silently skipped while every tool hits prod. Either drop the legacy spellings from the guard or fail loudly when they're set.

justinwlin:
Addressed the review: wizard key verification now uses REST v2 and RUNPOD_API_BASE_URL; retired REST variables no longer disable credential preflight. Added regression coverage and corrected distribution migration notes, parity documentation, and stale vendor/patch-script references.

lukepiette (approved):
Re-reviewed 1607b47. Wizard verification now hits `sdkBase()/v2/pods`, the legacy REST host resolvers are removed so they can no longer disable the credential preflight, the changeset now calls out the Dockerfile/Smithery/Cursor removals, and the parity manifest and .gitignore references are fixed. SDK is pinned to 0.1.1 in the lockfile. Still approved.

PR #93, opened 2026-09-14
[4/4] Add optional hosted feedback and private journal tools

justinwlin:
Add configuration-gated hosted feedback, questions, and private journals with a Convex storage sink. Scrub secrets, resolve read identity from the caller credential, and fail softly when storage is unavailable. Local stdio stays unchanged.

| Order | PR | Review focus |
| --- | --- | --- |
| 4 **(this PR)** | [#93](https://github.com/runpod/runpod-mcp/pull/93) | Hosted agent learning |

Merge **1 → 2 → 3 → 4**. This PR targets `review/specgen-03-surface`. After a squash merge, rebase/retarget the remaining stack before merging the next PR.

Validation: Type check, lint, build and generation check passed; all 366 tests passed, including credential scoping, scrub behavior, storage confirmation and malformed-request handling. Storage tests use local doubles; no production Convex deployment was performed.


src/alp/scrub.ts
```diff
@@ -0,0 +1,229 @@
...
+      (segment === 'key' && i > 0 && KEY_QUALIFIERS.has(segments[i - 1])) ||
+      // No separator at all: `apikey`, `accesskey`.
+      (segment.endsWith('key') && KEY_QUALIFIERS.has(segment.slice(0, -3)))
+  );
+}
+
+// Covers JSON, YAML and shell env assignments, including quoted values with
+// spaces or escaped quotes. The field name is kept for diagnostic context.
+const ASSIGNMENT =
+  /(?<![A-Za-z0-9_.-])(["']?[A-Za-z_][A-Za-z0-9_.-]*["']?\s*[:=]\s*)("(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*'|[^\s,;{}"']+)(?=\s|[,;}]|$)/g;
```
lukepiette, line 143: Reproduced: `db:\n  password: hunter2hunter2` and `environment:\n  RUNPOD_API_KEY: abcdef1234567890abcdef` both come back with 0 redactions. `\s*` after `[:=]` crosses the newline and the unquoted value class accepts `password:` as the value. Use `[ \t]*` after the separator and add a nested-YAML case to the test.

src/alp/scrub.ts
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// Bump when any stage's rules change so stored rows record which pass they got.
+export const SCRUB_VERSION = 3;
+
+const MARKER = /^["']?\[redacted:[a-z_]+\]["']?$/;
+
+// ---- Stage A: anchored credential shapes --------------------------------
+
+const CREDENTIAL_PATTERNS: Array<{ name: string; re: RegExp }> = [
```
lukepiette, line 33: No pattern for Hugging Face tokens (`\bhf_[A-Za-z0-9]{30,}\b`) or URL query secrets (`[?&](api_key|token|key|secret)=[^&\s]+`). Both pass through untouched today. Given the `HF_TOKEN` support just added elsewhere, HF tokens in particular will show up in journal entries.

src/specgen/tools/alp.ts
```diff
@@ -0,0 +1,319 @@
...
+    },
+  };
+
+  const saveToJournal: CuratedTool = {
+    // FAIL-SOFT applies to the argument gate too: a misnamed optional field
+    // is ignored, never a 400 that derails the task the agent is actually on.
+    lenientArguments: true,
+    name: 'save_to_journal',
+    description:
+      "Save something you learned about using Runpod that a future session would benefit from (e.g. 'image X needs CUDA 12.8', 'endpoint type cannot be changed after create'). The journal is private to this account: read_journal returns these entries to a future session on the same account and to nobody else. Runpod reviews entries in aggregate to improve the platform. Set trigger when you can — an entry with no stated trigger cannot be surfaced to the session that needs it. Never include API keys or secrets.",
```
lukepiette, line 224: "to nobody else" is a binding promise at write time. If the Learning Commons direction is publish-by-default with scrubbing and de-identification, entries saved under this wording can't be published later. Suggest softening to something like "may be shared with other Runpod users after scrubbing and de-identification; set visibility to private to keep it to this account" and threading `visibility` through the schema instead of hardcoding it.

src/alp/ingest.ts
```diff
@@ -0,0 +1,259 @@
+// ALP ingest endpoint: POST /api/alp/submit (see docs/agent-learning-protocol.md).
+//
+// The ONE write path for all three ALP routes on BOTH transports — the npm
```
lukepiette, line 3: Nit: stale. Stdio never gets these tools in this PR; hosted HTTP is the only caller.

lukepiette (commented):
Reviewed with a local checkout: tsc and eslint clean, ALP suites 31/31, 360 tests pass overall (same 6 local-only `findClaudeBinary` failures as main).

What I traced and found solid: read identity comes only from the server-side `myself { id }` probe and any client-supplied identity is ignored; the API key is never stored, only a sha256 hash for the verdict cache; `listJournalByIdentity` is an `internalQuery` on an index behind the shared secret and returns no attribution fields, so one account cannot read another's journal. Tools are appended only when both `ALP_SINK_URL` and `ALP_SINK_SECRET` are set on the HTTP path and stdio never passes them, so unset means absent from `tools/list`, not present-and-failing. Fail-soft is honest: sink down returns `recorded: false`, not a fake success. Admin purge is internal, dry-run by default.

Holding off on approval for one small code fix and one wording decision:

1. **Scrub misses nested YAML / compose blocks (reproduced).** `ASSIGNMENT` lets `\s*` after `[:=]` span a newline and indent, and the unquoted value class allows `:`, so for `db:\n  password: hunter2hunter2` it matches key=`db`, value=`password:` and swallows the child key. Zero redactions on that input and on `environment:\n  RUNPOD_API_KEY: ...`. Flat YAML works, which is all the current test covers. Fix is `[ \t]*` instead of `\s*` after the separator (and/or forbid a trailing `:` in the unquoted value) plus a nested case in the test. This is the most common paste shape for our users so I'd land it here.
2. **No Hugging Face token pattern.** `hf_[A-Za-z0-9]{30,}` passes through untouched, as do `?api_key=`/`?token=` URL query secrets. We just added `HF_TOKEN` support and agents paste these into env blocks constantly. One-line additions.
3. **Privacy wording vs the ALP direction.** The tool descriptions promise entries are visible "to nobody else", and the schema hardcodes `visibility: 'private'`. The ALP PRD has publish-by-default (anonymous, scrubbed) locked, with private as a per-entry override. Anything saved under a "nobody else" promise can't later be published without breaking it. I lean toward softening the description now ("may be published after scrubbing and de-identification; set visibility to private to opt out") rather than flipping the default later. Let's settle this in the thread before merge.

Nits inline: stale "BOTH transports" comment in `ingest.ts`, and the design doc says the ingest tree lives in `api/alp/` and calls the endpoint rate-limited, neither of which matches the code.

justinwlin:
Addressed the review: fixed nested YAML/Compose secret redaction, added Hugging Face token and URL query-secret redaction, and added regression/idempotency tests. Corrected hosted-only topology and rate-limit documentation. Journals remain private; public sharing will not be implemented. Wording now clearly discloses Runpod review access.

src/alp/scrub.ts
```diff
@@ -0,0 +1,250 @@
...
+      (segment === 'key' && i > 0 && KEY_QUALIFIERS.has(segments[i - 1])) ||
+      // No separator at all: `apikey`, `accesskey`.
+      (segment.endsWith('key') && KEY_QUALIFIERS.has(segment.slice(0, -3)))
+  );
+}
+
+// Covers JSON, YAML and shell env assignments, including quoted values with
+// spaces or escaped quotes. The field name is kept for diagnostic context.
+const ASSIGNMENT =
+  /(?<![A-Za-z0-9_.-])(["']?[A-Za-z_][A-Za-z0-9_.-]*["']?[ \t]*[:=][ \t]*)("(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*'|[^\s,;{}&#"']+)(?=\s|[,;}&#]|$)/g;
```
lukepiette: Regression from adding `&#` to the unquoted value class: `password: p&ss#word123` now becomes `password: [redacted:config]&ss#word123`, and `password=abc&def` leaks `&def`. Before this change both were fully redacted. The `&#` exclusion only matters for query strings, which stage C already handles, so one option is to keep the old value class here and make the `MARKER` check tolerate a `[redacted:query]` prefix instead. Add a `p&ss#word` case to the test either way.

lukepiette (commented):
Re-reviewed ce4d449. The code items are resolved and I reproduced each locally against the new scrubber:

- `db:\n  password: ...` and a Compose `environment:` block with `RUNPOD_API_KEY` and `HF_TOKEN` now redact fully.
- Bare Hugging Face tokens and `?api_key=` / `?token=` query values redact, with the non-secret query fields preserved.
- `tokenizer: llama-3` and a plain `url:` line are still left alone. Type-check clean, 367 tests pass (same 6 machine-local `findClaudeBinary` failures as main). Stale transport comments and the design doc are corrected.

One small regression inline: excluding `&` and `#` from the unquoted value class means an unquoted password containing either character is now only partially redacted.

The one item still open is the privacy wording. The new text ("entries are not published to other accounts") is a firm commitment, and the reply says public sharing will not be implemented. That is a product decision rather than a code one, so I'm holding approval until it's settled in this thread.

lukepiette (approved):
Confirmed: journal entries stay private to the account for now, so the wording is correct as written. Approving. The `&`/`#` partial-redaction nit above is still worth a quick follow-up but doesn't block merge.

